PR #68, opened 2026-03-16
fix: handle stale transaction in apply_filters

axisrow:
## Summary
- Fix `sqlite3.OperationalError: cannot start a transaction within a transaction` when clicking "Обновить статистику" (`POST /channels/filter/analyze`)
- A prior interrupted operation could leave an uncommitted transaction on the shared aiosqlite connection; now we catch the error, rollback the stale transaction, and retry `BEGIN`
- Atomicity of the reset+apply operation is preserved

## Test plan
- [x] All 33 filter tests pass (`test_filters.py`), including `test_apply_filters_is_atomic_on_error`
- [ ] Manual: click "Обновить статистику" on the channels page — no 500 error

### DIFF
--- a/src/filters/analyzer.py
+++ b/src/filters/analyzer.py
@@ -168,7 +168,9 @@ class ChannelAnalyzer:
         # (isolation_level=None means autocommit, so rollback is a no-op when clean).
         try:
             await conn.execute("BEGIN")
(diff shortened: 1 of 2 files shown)


axisrow:

PR #224, opened 2026-03-25
fix: default agent tool permissions to read-only

axisrow:
## Summary
- Change default tool permissions from all-enabled to read-only
- Write and delete tools are now disabled by default until explicitly enabled via Settings → Разрешения агента
- New tools added in future inherit the default for their category (read=on, write/delete=off)

## Test plan
- [ ] Fresh install (no `agent_tool_permissions` in DB) → only read tools available to agent
- [ ] Existing install with saved permissions → saved values preserved, no change
- [ ] Enable write/delete in UI → save → agent gains access to those tools
- [ ] `pytest tests/ -v -m "not aiosqlite_serial" -n auto` passes

### DIFF
--- a/src/agent/tools/permissions.py
+++ b/src/agent/tools/permissions.py
@@ -212,18 +212,27 @@ MODULE_GROUPS: OrderedDict[str, list[str]] = OrderedDict([
 # ---------------------------------------------------------------------------
 
 
+def _default_permissions() -> dict[str, bool]:
+    """Default permissions: read=True, write/delete=False."""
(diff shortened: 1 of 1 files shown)


axisrow:
Fixed in commit c02428f: changed all fallback defaults from `True` to `False` in `filter_allowed_tools` and `build_template_context` for deny-by-default at the enforcement boundary. While unreachable in current code (both lists are kept in sync at 93 tools), this hardens against future drift.

PR #275, opened 2026-03-27
feat: add Telegram search agent tools

axisrow:
## Summary
- Add 4 new agent tools for Telegram search modes (closes part of #274)
- `search_telegram` — global Premium search across public channels
- `search_my_chats` — search user's own dialogs
- `search_in_channel` — search within a specific channel via Telegram API
- `search_hybrid` — combined FTS + semantic local search

## Details
- Tools use lazy `SearchEngine` instance (created once on first call)
- `search_telegram`, `search_my_chats`, `search_in_channel` require `ClientPool` (guarded by `require_pool()`)
- `search_hybrid` works locally without pool
- All tools registered in `TOOL_CATEGORIES` and `MODULE_GROUPS` in `permissions.py`
- 11 new tests covering success, error, no-pool scenarios

## Test plan
- [x] `ruff check` passes
- [x] All 11 new tests pass
- [x] Existing 330 search-related tests still pass

### DIFF
--- a/src/agent/tools/permissions.py
+++ b/src/agent/tools/permissions.py
@@ -29,6 +29,10 @@ TOOL_CATEGORIES: dict[str, ToolCategory] = {
     "search_messages": ToolCategory.READ,
     "semantic_search": ToolCategory.READ,
     "index_messages": ToolCategory.WRITE,
+    "search_telegram": ToolCategory.READ,
+    "search_my_chats": ToolCategory.READ,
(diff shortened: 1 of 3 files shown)


axisrow:
**Re: channel_id=0 check** — SKIP. `channel_id=0` is never a valid Telegram ID (all IDs are positive/negative, never zero). The current `if not channel_id` guard is functionally correct for all real-world inputs. Changing to `is None` would only improve an impossible error message.

PR #282, opened 2026-03-28
fix: entity cache warming for numeric chat IDs + phone permissions default

axisrow:
## Summary
- **Entity resolution bug**: все 16 инструментов в `messaging.py` вызывали `client.get_entity()` без прогрева кэша — фейлились на числовых ID приватных супергрупп без username. Добавлен хелпер `resolve_entity()` в `_registry.py`, который роутит числовые ID через `ClientPool.resolve_dialog_entity()` (с `warm_dialog_cache()` fallback).
- **Phone permissions bug**: телефон, отсутствующий в perms dict, получал DENY вместо defaults (all enabled). Теперь запрещаются только телефоны с явным `tool=False`.

## Test plan
- [ ] `pytest tests/ -m "not aiosqlite_serial" -n auto` — 3922 passed
- [ ] `read_messages` с числовым ID приватной супергруппы (например, `1877929309`)
- [ ] `search_my_telegram` на аккаунте без явно настроенных разрешений

### DIFF
--- a/src/agent/tools/_registry.py
+++ b/src/agent/tools/_registry.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json as _json
 import logging
+import re as _re
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


axisrow:
**Comment 1 (permissions allowlist) — HALLUCINATION**

The reviewer's claim that this is a breaking security regression is not supported by the code.

Evidence from the codebase:

1. `save_tool_permissions()` has an explicit comment (line ~376): `# Seed unsaved accounts with all-enabled defaults so they are not implicitly denied`. The original developer's intent was always **default-allow** for phones not in the dict.

2. The old behaviour (deny phones absent from dict) was a **bug**: accounts added after the initial permission save were silently blocked. The user reported this exact issue with `search_my_telegram` being denied on a newly-added account.

3. The semantics remain strict for phones **explicitly present** with `tool=False` — those are still denied. Only phones **completely absent** from the dict get all-enabled defaults (consistent with what `save_tool_permissions` was already trying to guarantee via seeding).

No fix needed.

PR #306, opened 2026-03-29
fix(tui): toggle Send button to Stop during streaming

axisrow:
## Summary
- Кнопка "Отправить" теперь переключается на "■ Стоп (esc)" (красная) во время стриминга ответа агента
- Нажатие кнопки во время стриминга отменяет генерацию (аналог ESC)
- Кнопка гарантированно возвращается в исходное состояние через `finally` блок

## Test plan
- [ ] `ruff check src/cli/commands/agent_tui.py` — passes
- [ ] `pytest tests/test_agent_tui.py -v` — 28 tests pass
- [ ] Manual: `python -m src.main agent chat` → send message → button changes to red "■ Стоп (esc)" → on completion reverts to green "Отправить"
- [ ] Manual: click "■ Стоп (esc)" during streaming → cancels generation

### DIFF
--- a/src/cli/commands/agent_tui.py
+++ b/src/cli/commands/agent_tui.py
@@ -415,7 +415,10 @@ class AgentTuiApp(App):
 
     @on(Button.Pressed, "#send-btn")
     async def on_send_btn(self) -> None:
-        await self.action_send_message()
+        if self._stream_worker is not None and self._stream_worker.state in (WorkerState.PENDING, WorkerState.RUNNING):
+            self.action_cancel_stream()
+        else:
+            await self.action_send_message()
 
     async def action_new_thread(self) -> None:
         tid = await self.db.create_agent_thread("Новый тред")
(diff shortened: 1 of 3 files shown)


axisrow:
## Response to review comments

### Comment 1: Race condition — `HALLUCINATION`

The claim that `run_worker()` with `thread=False` could let the worker's `finally` block execute before `_set_button_streaming(True)` is **incorrect**.

With `thread=False`, the worker is an **async task** on the same event loop. Python's asyncio is cooperative (not preemptive) — the current coroutine continues executing until it yields (`await`). Since `_set_button_streaming(True)` is a synchronous call, it **always completes before the worker starts**.

**Test proof** (`test_no_race_condition_button_set_before_worker_runs`):
- Uses an instant-completing stream (worst case scenario)
- Tracks call order via monkeypatch
- Asserts `set_button(True)` always precedes `set_button(False)`
- **Result: PASSED** — no race condition even in the degenerate case

Textual docs confirm: `run_worker` returns "almost immediately" — the coroutine is *scheduled*, not *started*.

### Comment 2: `query_one` in `finally` during teardown — `SKIP`

Theoretical edge case. In practice, Textual cancels workers and processes the `finally` block while the widget tree is still mounted. The `query_one` call succeeds because worker cancellation happens before widget dismount in Textual's shutdown sequence.

Added 4 verification tests in commit 169a9d4:
- `test_button_toggles_to_stop_during_streaming` ✅
- `test_button_reverts_after_cancel` ✅  
- `test_stop_button_click_cancels_streaming` ✅
- `test_no_race_condition_button_set_before_worker_runs` ✅

PR #340, opened 2026-04-02
fix(filters): check stats before filtering to avoid strange symbols

### LINKED ISSUE #331: Strange sympols on the Clearning tab
Image

To reproduce:
Click on "Refresh" without collecting data

Link: /channels/filter/manage

axisrow:
## Summary
- Add GET `/channels/filter/has-stats` endpoint to check whether active channels already have statistics
- Replace the filter refresh form with JavaScript that probes stats first and keeps `refreshFilters()` available even in the empty state
- If stats are missing, queue the existing `/channels/stats/all` flow and clearly tell the user to return and rerun filters after stats finish
- Keep `/channels/filter/analyze` focused on filter analysis only, without inline stats collection inside the HTTP request

## Test plan
1. Start server: `python -m src.main serve`
2. Add a channel without collecting stats
3. Open `/channels/filter/manage`
4. Click "Обновить фильтры" → confirm dialog explains that only stats collection runs now
5. Confirm → stats collection is queued and the user is redirected to `/channels`
6. After stats finish, return to `/channels/filter/manage` and click "Обновить фильтры" again
7. With stats present, filters apply immediately without the confirm dialog

Fixes #331

### DIFF
--- a/src/web/routes/filter.py
+++ b/src/web/routes/filter.py
@@ -115,8 +115,13 @@ async def hard_delete_selected(request: Request):
 
 
 @router.post("/filter/analyze")
-async def analyze_channels(request: Request):
+async def analyze_channels(request: Request, with_stats: bool = False):
     db = deps.get_db(request)
+
+    if with_stats:
+        svc = deps.collection_service(request)
+        await svc.collect_all_stats()
+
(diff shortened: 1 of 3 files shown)


axisrow:

PR #359, opened 2026-04-05
fix(scheduler): single-row progress counter for stats update task

axisrow:
## Summary
- Replaces continuation-task pattern (new DB row every 20 channels) with single-row progress tracking for "Обновление статистики" — counter now shows 1/234 → 234/234 on one row instead of spawning ~12 separate rows
- Flood-wait reschedules the same task back to PENDING instead of creating a continuation; cancel from UI takes effect within one channel cycle
- Counter now tracks processed channels (ok+err) so it always reaches total, even when some channels fail

## Test plan
- [x] All 5160 tests pass (4489 parallel + 671 serial)
- [ ] Manual: start server, trigger "Обновить статистику всех", verify single row on /scheduler with incrementing counter
- [ ] Manual: press "Отменить" mid-run, verify task stops within seconds
- [ ] Manual (optional): trigger flood-wait, verify task returns to PENDING and resumes automatically

### DIFF
--- a/src/agent/tools/collection.py
+++ b/src/agent/tools/collection.py
@@ -88,7 +88,7 @@ def register(db, client_pool, embedding_service, **kwargs):
                 return _text_response(f"Канал pk={pk} не найден.")
             from src.models import StatsAllTaskPayload
 
-            payload = StatsAllTaskPayload(channel_ids=[ch.channel_id], batch_size=1)
+            payload = StatsAllTaskPayload(channel_ids=[ch.channel_id])
             await db.create_stats_task(payload)
             return _text_response(f"Сбор статистики для '{ch.title}' поставлен в очередь.")
         except Exception as e:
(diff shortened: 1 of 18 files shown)


axisrow:
All three issues from the review have been addressed in commit a29ca52:

**Bug 1 (crash recovery):** Added `persist_stats_progress()` which atomically saves `next_index`, `channels_ok`, `channels_err`, and `messages_collected` to the DB row on every channel iteration. On crash, startup recovery now resumes from the correct cursor instead of restarting from 0.

**Bug 2 (graceful shutdown marking COMPLETED):** After `stop_event` interruption, the code now calls `reschedule_stats_task` with the current cursor (status → PENDING) instead of unconditionally marking COMPLETED. Remaining channels will be processed on next startup.

**Issue 3 (completed_at not cleared):** Added `completed_at = NULL` to the `reschedule_stats_task` SQL UPDATE as a defensive guard.

PR #394, opened 2026-04-06
fix(web): pre-render templates to prevent "No response returned" in middleware stack

axisrow:
## Summary

- Pre-render scheduler template synchronously in handler so Jinja2 errors surface before the ASGI response phase, preventing `RuntimeError("No response returned.")` from 5-layer `BaseHTTPMiddleware` stack
- Add try/except diagnostic logging in `scheduler_page` to capture the real exception before middleware masks it
- Pre-render `error.html` in `unhandled_exception_handler` to prevent double-swallow when the error page itself fails to render
- Add 2 regression tests covering the trigger-collection→redirect→render flow

Temporary workaround until full ASGI middleware migration in #393.

## Test plan

- [x] All 45 scheduler tests pass
- [x] Lint clean (`ruff check`)
- [ ] Manual: Start scheduler → Collect all channels → page renders without 500

Closes #393

### DIFF
--- a/src/web/app.py
+++ b/src/web/app.py
@@ -290,15 +290,13 @@ def create_app(config: AppConfig | None = None) -> FastAPI:
             )
 
         try:
-            return app.state.templates.TemplateResponse(
-                request,
-                "error.html",
-                {
-                    "status_code": 500,
-                    "detail": "An unexpected error occurred. See /debug/ for details.",
-                },
-                status_code=500,
-            )
+            tpl = app.state.templates.env.get_template("error.html")
+            body = tpl.render({
+                "request": request,
+                "status_code": 500,
+                "detail": "An unexpected error occurred. See /debug/ for details.",
+            })
+            return HTMLResponse(body, status_code=500)
         except Exception:
             return HTMLResponse("Internal Server Error", status_code=500)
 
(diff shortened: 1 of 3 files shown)


axisrow:
## Triage of Claude Review Comments

All three review comments have been verified against the source code. Verdict: **no fixes required**.

### 1. Double-logging on scheduler exceptions — **SKIP**
The `logger.exception("scheduler_page failed")` at `scheduler.py:116` followed by `raise` intentionally captures the original stack trace before middleware can mask it. The duplicate log entry from `unhandled_exception_handler` in `app.py:284` is a known, acceptable tradeoff for this workaround. This is an admin-only page — the operational impact of double-logged entries is minimal and does not constitute a bug or data risk.

### 2. Synchronous `tpl.render()` blocks event loop — **SKIP**
The reviewer explicitly acknowledges this is a "deliberate tradeoff of the workaround" and "acceptable for an admin panel." Jinja2 rendering of ~50-100 task rows takes milliseconds — not a meaningful production concern for an internal tool. This is not a performance regression in practice.

### 3. `dry_run_notifications` still uses `TemplateResponse` — **SKIP**
The reviewer confirms there is "no masking risk" because the endpoint is HTMX-only and `unhandled_exception_handler` returns a plain `<div>` for HTMX requests without attempting template rendering. No fix needed — the asymmetry is intentional and correct.

**Conclusion:** The review found no critical bugs, security vulnerabilities, or data loss risks. All comments are advisory observations about known tradeoffs of the workaround approach. No code changes are warranted.

PR #443, opened 2026-04-14
fix(collector): select correct account for private groups

### LINKED ISSUE #442: Private group collection fails: wrong Telegram account selected, no test coverage for multi-account entity resolution
## Bug

Collecting messages from a private Telegram group (supergroup without username) fails with:

```
Could not find the input entity for PeerChannel(channel_id=XXXXXXXXX)
```

### Root cause

Telethon requires an `access_hash` to resolve a `PeerChannel`. The hash is only available in the entity cache of an account that is **a member of the group**. It gets populated via `get_dialogs()`.

The bug has two layers:

1. **Wrong account selected**: `get_available_client()` picks any available phone arbitrarily. In a multi-account setup, the selected phone may not be a member of the private group — the `access_hash` is then absent and resolution fails.

2. **Cache not warmed at startup**: `get_dialogs()` was called lazily (once per phone, per process, on first collection). After a restart the cache is empty until the first collection attempt for that phone.

### Fix (in this PR)

- `warm_all_dialogs()` runs at startup (background task) and on a 24 h scheduler job — populates Telethon's entity cache **for every connected account** and persists a `preferred_phone` field to the `channels` table.
- Collection of no-username channels now uses the lookup chain: `channels.preferred_phone` (DB, survives restarts) → in-memory map → wait for ongoing warm → `get_available_client()` fallback.
- On `ValueError` (stale cached phone, e.g. account was kicked): invalidates cache, rediscovers the correct account across all phones, persists the new mapping, retries. If no account has access — raises, error shown in UI.
- `warm_all_dialogs` is visible in the Scheduler UI with manual trigger and configurable interval.

---

## Test coverage gap

All existing tests for private-group / no-username collection used **a single mock phone**. In that setup the bug manifests as "entity not found → error shown" which is actually the *correct* outcome when no account has access — so tests passed.

The critical multi-account scenario was never exercised:

> **Phone A** is not a member → `ValueError`. **Phone B** is a member → should succeed.

### What needs to be addressed

- [ ] Add tests for multi-account entity resolution in similar flows (stats collection, dialog listing, any other place that calls `get_entity` / `resolve_entity` on a no-username peer).
- [ ] Audit all `get_available_client()` call-sites for operations that require the caller to be a group/channel member — each of those is a potential silent failure in multi-account setups.
- [ ] Consider a fixture-level `FakeClientPool` upgrade that makes multi-phone scenarios easy to set up, to prevent regression of this class of bug.

axisrow:
Closes #442

## What changed

Collection of private Telegram groups (supergroups without username) picks the account that actually has access, instead of failing with `Could not find the input entity for PeerChannel(...)`.

- **`channels.preferred_phone`** — new DB column; persists `channel_id → phone` across restarts.
- **`warm_all_dialogs()`** — prefills Telethon's entity cache for **every connected account** and populates `preferred_phone`. Runs at startup (background task) and on a scheduler job (default: every 24h).
- **Lookup chain for no-username channels**: `channels.preferred_phone` → in-memory map → wait for ongoing warm (with 30s timeout) → fallback to `get_available_client()`.
- **Error recovery**: on `ValueError` from `resolve_entity` (stale phone, account was kicked, etc.) — invalidate cached phone, rediscover across remaining phones, persist new mapping, retry. If no account has access — raises → error visible in UI.
- **Scheduler UI**: `warm_all_dialogs` shows up alongside other jobs with "Запустить сейчас" trigger and configurable interval.

## Why

`get_available_client()` picks any available phone arbitrarily. In multi-account setups the selected phone may not be a member of the private group — its entity cache has no `access_hash` for that channel and Telethon fails the resolve. Users saw "Ошибка" in the UI for every sync of a private group that the pool happened to route through the wrong account.

## Root cause

Two layers:

1. **Wrong account selected** — no routing by membership.
2. **Cache cold after restart** — dialogs were fetched lazily (once per phone, per process) on first collection attempt, so a fresh start left the cache empty.

Both are fixed: account is explicitly chosen by DB mapping, and the cache is warmed up-front for all accounts.

## Impact

- End users: collection of private groups now works as long as at least one connected account is a member.
- DB: new nullable `preferred_phone` column; migration is additive (`ALTER TABLE ADD COLUMN`), no downtime.
- Scheduler: one new visible job; disabled users can toggle or adjust interval.
- Behaviour unchanged for public channels (username path).

## Test coverage gap

The original single-phone tests passed because in a single-phone setup "entity not found → error" is the correct outcome. The bug only manifests in multi-phone setups, which no test covered.

This PR adds:

- `test_collect_private_group_uses_map_phone` — map hit, uses the phone directly (no fallback).
- `test_collect_private_group_discovers_access_phone` — two phones, first fails, second succeeds; map + DB get updated.

See #442 for the broader audit of `get_available_client()` call sites and fixture-level improvements.

## Validation

- `ruff check src/ tests/` — clean.
- `pytest tests/ -v -m "not aiosqlite_serial" -n auto` — **4938 passed, 1 skipped**.

### DIFF
--- a/src/database/facade.py
+++ b/src/database/facade.py
@@ -277,6 +277,12 @@ class Database:
         self._require()
         await self._channels.set_channel_type(channel_id, channel_type)
 
+    async def update_channel_preferred_phone(
+        self, channel_id: int, phone: str | None
+    ) -> None:
+        self._require()
+        await self._channels.update_channel_preferred_phone(channel_id, phone)
+
     async def update_channel_meta(
         self, channel_id: int, *, username: str | None, title: str | None
     ) -> None:
--- a/src/database/migrations.py
+++ b/src/database/migrations.py
@@ -120,6 +120,9 @@ async def run_migrations(db: aiosqlite.Connection) -> bool:
     if "created_at" not in ch_columns:
         await db.execute("ALTER TABLE channels ADD COLUMN created_at TEXT")
         await db.commit()
+    if "preferred_phone" not in ch_columns:
+        await db.execute("ALTER TABLE channels ADD COLUMN preferred_phone TEXT")
+        await db.commit()
 
     cur = await db.execute("PRAGMA table_info(collection_tasks)")
     task_rows = await cur.fetchall()
--- a/src/database/repositories/channels.py
+++ b/src/database/repositories/channels.py
@@ -213,6 +213,14 @@ class ChannelsRepository:
         )
         await self._db.commit()
 
+    async def get_preferred_phone(self, channel_id: int) -> str | None:
+        """Return the preferred phone for a channel, or None if not set."""
+        row = await self._db.fetchone(
+            "SELECT preferred_phone FROM channels WHERE channel_id = ?",
+            (channel_id,),
+        )
+        return row["preferred_phone"] if row else None
+
     async def update_channel_created_at(self, channel_id: int, created_at) -> None:
         """Set created_at only if currently NULL (backfill from entity.date)."""
         iso = created_at.isoformat() if hasattr(created_at, "isoformat") else created_at
--- a/src/database/repositories/channels.py
+++ b/src/database/repositories/channels.py
@@ -69,6 +69,9 @@ class ChannelsRepository:
                 if "message_count" in keys and row["message_count"] is not None
                 else 0
             ),
+            preferred_phone=(
+                row["preferred_phone"] if "preferred_phone" in keys else None
+            ),
         )
 
     async def get_channels(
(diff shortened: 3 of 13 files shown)


axisrow:

axisrow:
### Re: `_warming_task` set twice on startup path

Acknowledged. The double-assignment is harmless — both set the same `asyncio.Task` object. The bootstrap assignment exists so `is_warming()` returns `True` immediately after task creation, before the coroutine's first instruction executes. This is defensive programming, not a bug. Not changing in this PR.

axisrow:

axisrow:
### Re: warm_all_dialogs FloodWait in outer except

Acknowledged. Adding `FloodWaitError` catch in the outer exception handler of `warm_all_dialogs` is a good consistency improvement. The practical risk is low as noted — the phone is released and subsequent collection uses `run_with_flood_wait`. Will address in a follow-up if needed.

PR #454, opened 2026-04-17
feat: finish web-mode cutover for accounts and /agent (#453)

### LINKED ISSUE #453: Follow-up: finish web-mode cutover for accounts and agent after #444
Follow-up from #444.

## Summary
`#444` mostly completed the worker/web split, but two production-path gaps remain:

1. Account-management web routes still mutate Telegram runtime through `AccountService`, which breaks the rule that web must not touch live Telegram runtime.
2. `/agent` web UI still depends on an in-process `AgentManager`, but real web bootstrap does not create one outside worker mode.

This issue closes those gaps and adds regression coverage against the real web container wiring rather than the current live-harness test setup.

## Required changes
- Convert account-management actions to worker-owned flow.
  - Replace direct runtime mutations in the web request path with durable commands or purely DB-backed state changes.
  - `toggle` and `delete` account actions must no longer call `pool.add_client(...)` / `pool.remove_client(...)` from web-side services.
  - Define worker behavior for account activation, deactivation, and deletion so runtime state converges from DB truth.
- Finish `/agent` compatibility story for web mode.
  - Either disable `/agent` interaction in web mode with a clear and tested contract, or route agent execution through a worker-owned execution path.
  - Remove the current partial state where the UI exists but chat and permission endpoints fail with `503 AgentManager not initialized`.
- Tighten runtime guard coverage.
  - Extend guard tests to cover web-consumed services outside `src/web`, especially account-management paths.
  - Ensure guard coverage catches direct runtime calls and attribute-based runtime coupling.
- Add production-like web-mode tests.
  - Build relevant route tests through the real web container/bootstrap snapshot wiring, not the live harness pool/collector/scheduler shortcut.
  - Add regressions proving account routes and agent routes behave correctly when `runtime_mode="web"` and only snapshot/shim objects are present.

## Acceptance criteria
- Account POST actions keep existing URLs but no longer require live `ClientPool` in web mode.
- `/agent` behavior in web mode is explicit, deterministic, and covered by tests.
- No web route or web-consumed service requires live `ClientPool`, `Collector`, or live `AgentManager` in production web mode.
- Guard tests fail if account-management or agent-related web paths reintroduce live runtime access.

## Notes
Validated locally before filing:
- `./.venv/bin/python -m pytest tests/test_web_runtime_guards.py tests/routes/test_accounts_routes.py tests/routes/test_agent_routes.py -q`
- `./.venv/bin/python -m pytest tests/routes/test_auth_routes.py tests/routes/test_scheduler_routes.py tests/routes/test_scheduler_routes_extra.py tests/test_web_auth.py -q`

Parent issue: #444

---

### Concrete runtime breakage for account routes in web mode
`AccountService.toggle`/`delete` in `src/services/account_service.py:30,34,42` call `self._pool.add_client(...)` and `self._pool.remove_client(...)`. In web mode `deps.account_service()` (`src/web/deps.py:283`) is constructed with `get_pool(request)`, which returns `SnapshotClientPool` from `src/web/runtime_shims.py`. That shim intentionally does **not** implement `add_client` / `remove_client`, so the current behavior of `POST /accounts/{id}/toggle` and `POST /accounts/{id}/delete` in a real web bootstrap is an unhandled `AttributeError` (HTTP 500), not just a latent coupling. Acceptance tests must exercise both endpoints against the snapshot-shim wiring, not the live harness.

`POST /accounts/{id}/flood-clear` is already DB-only and is not affected. `src/cli/commands/account.py:76` still calls `pool.add_client(...)` directly — that is a CLI-only path and must be kept as-is (CLI runs inside worker-equivalent runtime); it is explicitly out of scope for this follow-up.

### Proposed command surface for the cutover
Introduce dedicated durable commands and make web routes pure producers:
- `accounts.toggle` — payload `{account_id}`; worker flips `is_active` and reconciles pool membership (`add_client`/`remove_client`) from DB truth.
- `accounts.delete` — payload `{account_id}`; worker detaches the client first, then deletes the DB row.

The existing `accounts.connect` handler (`src/services/telegram_command_dispatcher.py:520`) is the precedent to follow. Web should return 303 with `?command_id=…` (consistent with dialogs/channels cutover).

### Related scheduler bug in the same surface
Not strictly account/agent, but the same class of regression:
- `POST /scheduler/start` (`src/web/routes/scheduler.py:415`) calls `SnapshotSchedulerManager.start()` in web mode, which raises `RuntimeError("Scheduler runtime is only available in the worker process.")` → HTTP 500.
- Either gate registration of the scheduler router behind `runtime_mode == "worker"` in `src/web/assembly.py:178`, or return a clean 409/400 from the route handler. `stop()` is already a no-op and is safe.

### Guard tests miss attribute-based coupling
`tests/test_web_runtime_guards.py` only checks for method calls like `pool.get_available_client(` or `pool.get_client_by_phone(`. It does not flag attribute reads such as `pool.clients`, `pool._dialogs_cache`, `pool._active_leases`, `pool._premium_flood_wait_until`, which currently appear in `src/web/routes/debug.py`, `src/web/routes/settings.py:459`, `src/web/routes/photo_loader.py:246–248`, and `src/web/routes/dashboard.py:46,54`. Today these are safe because `SnapshotClientPool` exposes a read-only `clients` property and the debug route already uses `getattr(..., default)` for private attrs, but the guard should explicitly lock that contract down (allow-list of attribute names, reject anything outside it) so a future direct-mutation regression is caught.

### Missing offline-startup regression test from #444 Test Plan
The #444 Test Plan requires "web startup without internet does not initiate a single Telegram connect/reconnect and `/scheduler`, `/settings`, `/debug` open successfully." This is achieved architecturally via the snapshot shims, but there is no explicit regression test for it. Adding one here (or in a sibling follow-up) would prevent silent re-introduction of a live connect in the web bootstrap path.

axisrow:
## Summary

Closes #453. Finishes the web/worker cutover started in #444 for the two
remaining web-mode production-path gaps.

- **Accounts** — web `POST /settings/{id}/toggle` and `/delete` now
  enqueue `accounts.toggle` / `accounts.delete` commands; the worker
  flips `is_active` (or deletes the row) and reconciles pool membership
  from DB truth. `AccountService` stays in the repo for worker/CLI
  callers but is no longer wired into the web container.
- **/agent in web mode** — explicit 503 contract with actionable
  message (`Agent chat is only available in the worker process.
  Run \`python -m src.main worker\` alongside the web server.`) plus a
  banner on `GET /agent` instead of the old opaque "AgentManager not
  initialized".
- **Guard tests** — extended to cover mutating attribute access on
  `pool.clients`, direct private-attr reads, and a new check over
  web-consumed services outside `src/web/` (`filter_deletion_service`,
  `notification_service`, `notification_target_service`,
  `telegram_command_service`). A dedicated test blocks any re-import
  of `AccountService` into `src/web/`.
- **Offline-startup regression** — `test_build_web_container_does_not_connect_telethon`
  patches `TelegramClient.connect`, builds a real web container, and
  asserts no Telethon connect happens. Covers the #444 Test Plan item.
- **Production-like web-mode route tests** — new
  `tests/routes/test_accounts_agent_web_mode.py` exercises accounts and
  agent endpoints through the real `build_web_container` wiring
  (SnapshotClientPool/SnapshotCollector/SnapshotSchedulerManager, no
  `MagicMock` pool).

## Deliberate non-goals

- `channel_service.py` / `collection_service.py` are genuinely hybrid
  (some methods used by web, others only by worker) and are deferred
  from the guard list. Converting their remaining pool-touching paths
  to queued commands is a separate follow-up.
- `src/cli/commands/account.py:76` still calls `pool.add_client`
  directly. CLI runs in a worker-equivalent runtime, so this is correct
  and explicitly out of scope for #453.

## Test plan

- [x] `ruff check src/ tests/ conftest.py`
- [x] `pytest tests/ -q -m "not aiosqlite_serial" -n auto` — 5859 passed
- [x] `pytest tests/ -q -m aiosqlite_serial` — 741 passed
- [x] New tests:
  - `tests/routes/test_accounts_agent_web_mode.py::*` (6 tests, real web bootstrap)
  - `tests/test_bootstrap.py::test_build_web_container_does_not_connect_telethon`
  - `tests/test_web_runtime_guards.py::test_web_consumed_services_do_not_call_live_pool_methods`
  - `tests/test_web_runtime_guards.py::test_account_service_not_imported_by_web`
- [ ] Manual: GET /agent in web mode shows the yellow banner and POSTs to
  chat/permission return `{"detail": "Agent chat is only available in the
  worker process..."}` (503).

### DIFF
--- a/src/services/telegram_command_dispatcher.py
+++ b/src/services/telegram_command_dispatcher.py
@@ -548,6 +548,35 @@ class TelegramCommandDispatcher:
         await self._db.update_account_premium(phone, is_premium)
         return {"phone": phone, "is_premium": is_premium}
 
+    async def _handle_accounts_toggle(self, payload: dict[str, Any]) -> dict[str, Any]:
+        account_id = int(payload["account_id"])
+        accounts = await self._db.get_accounts()
+        account = next((a for a in accounts if a.id == account_id), None)
+        if account is None:
+            raise RuntimeError(f"account_not_found:{account_id}")
+        new_active = not account.is_active
+        await self._db.set_account_active(account_id, new_active)
+        if new_active:
+            try:
+                await self._pool.add_client(account.phone, account.session_string)
+            except Exception as exc:
+                logger.warning("accounts.toggle: failed to add client %s: %s", account.phone, exc)
+        else:
+            await self._pool.remove_client(account.phone)
+        return {"account_id": account_id, "is_active": new_active}
+
+    async def _handle_accounts_delete(self, payload: dict[str, Any]) -> dict[str, Any]:
+        account_id = int(payload["account_id"])
+        accounts = await self._db.get_accounts()
+        account = next((a for a in accounts if a.id == account_id), None)
+        if account is not None:
+            try:
+                await self._pool.remove_client(account.phone)
+            except Exception as exc:
+                logger.warning("accounts.delete: failed to remove client %s: %s", account.phone, exc)
+        await self._db.delete_account(account_id)
+        return {"account_id": account_id, "deleted": True}
+
     async def _handle_notifications_setup_bot(self, payload: dict[str, Any]) -> dict[str, Any]:
         bot = await self._notification_service().setup_bot()
         return {"bot_username": bot.bot_username, "bot_id": bot.bot_id}
--- a/src/services/telegram_command_dispatcher.py
+++ b/src/services/telegram_command_dispatcher.py
@@ -562,7 +562,10 @@ class TelegramCommandDispatcher:
             except Exception as exc:
                 logger.warning("accounts.toggle: failed to add client %s: %s", account.phone, exc)
         else:
-            await self._pool.remove_client(account.phone)
+            try:
+                await self._pool.remove_client(account.phone)
+            except Exception as exc:
+                logger.warning("accounts.toggle: failed to remove client %s: %s", account.phone, exc)
         return {"account_id": account_id, "is_active": new_active}
 
     async def _handle_accounts_delete(self, payload: dict[str, Any]) -> dict[str, Any]:
--- a/src/web/deps.py
+++ b/src/web/deps.py
@@ -23,7 +23,6 @@ from src.database.bundles import (
 from src.scheduler.service import SchedulerManager
 from src.search.ai_search import AISearchEngine
 from src.search.engine import SearchEngine
-from src.services.account_service import AccountService
 from src.services.channel_service import ChannelService
 from src.services.collection_service import CollectionService
 from src.services.filter_deletion_service import FilterDeletionService
@@ -273,14 +272,6 @@ def channel_service(request: Request) -> ChannelService:
         request,
         "_channel_service",
         lambda: ChannelService(get_db(request), get_pool(request), get_queue(request)),
-    )
-
-
-def account_service(request: Request) -> AccountService:
-    return _request_cached(
-        request,
-        "_account_service",
-        lambda: AccountService(get_account_bundle(request), get_pool(request)),
     )
 
 
--- a/src/web/routes/accounts.py
+++ b/src/web/routes/accounts.py
@@ -12,14 +12,28 @@ router = APIRouter()
 
 @router.post("/{account_id}/toggle")
 async def toggle_account(request: Request, account_id: int):
-    await deps.account_service(request).toggle(account_id)
-    return RedirectResponse(url="/settings?msg=account_toggled", status_code=303)
+    command_id = await deps.telegram_command_service(request).enqueue(
+        "accounts.toggle",
+        payload={"account_id": account_id},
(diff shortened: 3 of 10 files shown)


axisrow:

axisrow:
## Review triage after commit 412a7a5

**FIX:**
- `_handle_accounts_toggle` — wrapped the `pool.remove_client` call in `try/except Exception` to match `_handle_accounts_delete`. Prevents a FAILED command from leaving DB `is_active=False` while the pool client is still live.

**SKIP (cosmetic):**
- `client._transport.app.state.db` in `tests/routes/test_settings_routes.py` — this pattern is already used widely across the existing test suite (grep shows 30+ occurrences). Migrating all of them off `_transport` is worthwhile but is a broader test-refactor task, not scoped to #453.

axisrow:

PR #494, opened 2026-04-27
fix(scheduler): periodic DB pull picks up freshly enqueued PENDING tasks

axisrow:
## Symptom

Clicking **«Собрать все каналы»** on http://0.0.0.0:8080/scheduler/ — request returns 303 with \`collect_all_queued\` and the badge updates, but the worker never starts collecting. Tasks pile up as PENDING in \`collection_tasks\` and only get picked up after a worker restart.

## Root cause

In split / embedded-worker setups (web container has \`collection_queue=None\` because it uses snapshot shims), \`CollectionService._enqueue_channel\` falls back to a DB-only insert (\`src/services/collection_service.py:48-58\`). The worker's \`CollectionQueue\` only runs \`requeue_startup_tasks\` **once**, in \`bootstrap.py:333\`. There is no periodic re-check, so a PENDING row written after worker startup is invisible to the in-memory queue forever.

\`UnifiedDispatcher\` polls the DB every 5s but explicitly skips \`channel_collect\` (\`src/services/unified_dispatcher.py:46\` — _\"Polls DB for non-CHANNEL_COLLECT tasks\"_), so it doesn't cover this either.

## Fix

Add a periodic DB-pull loop to \`CollectionQueue\` (\`DB_PULL_INTERVAL_SEC=3.0\`):

- Extracted the existing requeue body into \`_ingest_pending_tasks()\`, used by both the startup path (\`requeue_startup_tasks\`) and the new \`_db_pull_loop\`.
- Added \`_known_task_ids: set[int]\` to dedupe against tasks already sitting in the in-memory queue or in delayed-requeue, so the pull loop never double-ingests.
- \`start_db_pull()\` is called from \`bootstrap.start_container\` right after \`requeue_startup_tasks\`; \`shutdown()\` stops the loop cleanly.
- The loop swallows exceptions per iteration with \`logger.exception\` — never crashes the worker.

## Closes

Closes the \"trigger collection does nothing\" symptom that has surfaced repeatedly since the worker/web split (#444 / #457). The existing e2e \`test_trigger_collection_persists_messages_via_embedded_worker\` was passing only by luck (web wrote PENDING before the worker's startup requeue ran); now the pickup is guaranteed.

## Test plan

- [x] 4 new unit tests in \`tests/test_collection_queue_db_pull.py\` (pickup, no double-ingest, error swallowing, idempotent stop).
- [x] Existing \`test_collection_flow.py\` e2e suite still green (3 tests).
- [x] Full suite: 7535 passed, 1 skipped.
- [x] \`ruff check\` clean.
- [ ] Manual: click \"Собрать все каналы\" with embedded worker, watch logs — collection should start within ~3 s.

### DIFF
--- a/src/collection_queue.py
+++ b/src/collection_queue.py
@@ -323,7 +323,7 @@ class CollectionQueue:
         except asyncio.QueueFull:
             logger.warning(
                 "Collection queue full on reconnect requeue; task %d stays PENDING "
-                "and will be picked up by requeue_startup_tasks",
+                "and will be picked up by the DB pull loop",
                 task_id,
             )
             return False
--- a/src/collection_queue.py
+++ b/src/collection_queue.py
@@ -94,6 +94,7 @@ class CollectionQueue:
         for task in list(self._delayed_requeues):
             task.cancel()
         self._delayed_requeues.clear()
+        self._known_task_ids.clear()
         logger.info(
             "Cleared %d pending collection tasks from DB and %d queued items from memory",
             deleted,
@@ -124,9 +125,10 @@ class CollectionQueue:
             except asyncio.QueueFull:
                 logger.warning(
                     "Collection queue full on delayed requeue; task %d stays PENDING "
-                    "in DB and will be picked up by requeue_startup_tasks",
+                    "in DB and will be picked up by the DB pull loop",
                     task_id,
                 )
+                self._known_task_ids.discard(task_id)
             self._ensure_worker()
 
         # Reserve the slot up front so the periodic DB pull does not double-ingest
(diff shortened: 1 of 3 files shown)


axisrow:

axisrow:

axisrow:
Addressed both queue-id lifecycle findings in e9cb19f: delayed requeue QueueFull now discards the reserved task id so the DB pull loop can retry it, and clear_pending_tasks() clears known task ids after draining/cancelling pending in-memory work. Added regression coverage for both paths. Validation: ruff check src/ tests/; python3 -m pytest tests/test_collection_queue_db_pull.py -v; python3 -m pytest tests/ -v (7537 passed, 1 skipped).

PR #502, opened 2026-04-28
fix(collector): throttle username resolves on long FloodWait

### LINKED ISSUE #464: Collector resolve_username without rate-limit triggers 16h flood-wait on all accounts
## Problem

Action-only pipelines (e.g. pipeline 1 «Реакции @claudecodeteam») rely on `ClientPool.get_available_client()` returning a live client. But all three accounts in the pool regularly end up simultaneously flood-waited for 12-16 hours, which makes every scheduled pipeline run return ${\tt result\_count=0}$ and forcibly skips every action.

Observed in \`data/app.log\` on 2026-04-18 22:43:55 — both primary and secondary accounts got 57891s flood-wait back-to-back within minutes:

\`\`\`
22:43:55 [WARNING] src.telegram.client_pool: Flood wait for +66982102247: 57891 seconds (until 2026-04-19 06:48:46 UTC)
22:43:55 [WARNING] src.telegram.collector: collect_channel_resolve_username: Flood wait 57891s until 2026-04-19T06:48:46 UTC for +66982102247
\`\`\`

All three accounts (+66982102247, +66990712629, +8613392919509) were flood-locked in the same run, ruining pipeline execution for the next 12-16 hours.

## Root cause

\`src/telegram/collector.py:546-565\` — \`_collect_channel\` calls \`session.resolve_entity(channel.username)\` for every channel in a loop. On large channel lists this blast triggers Telegram's \`auth.resolveUsername\` rate-limit, which is account-level and yields multi-hour flood-waits.

\`\`\`python
try:
    entity = await run_with_flood_wait(
        session.resolve_entity(channel.username),
        operation=\"collect_channel_resolve_username\",
        phone=phone,
        pool=self._pool,
        ...
    )
except HandledFloodWaitError as exc:
    flood_wait_sec = exc.info.wait_seconds
    raise  # ← propagates up, collector keeps dispatching to other accounts
\`\`\`

Because the collector keeps trying further channels on the *next* account after catching flood, it walks through the whole pool and poisons every account within a single cycle.

## Impact

- Action pipelines (react/forward/delete) cannot run at all while pool is flood-locked.
- Observable symptom is a silent \`result_count=0\` with no user-visible reason (fixed separately via node-level \`metadata.node_errors\` in follow-up to issue #463).
- Scheduler keeps enqueueing PIPELINE_RUN tasks every 5 minutes; all of them noop.

## Proposed fix

1. **Rate-limit \`resolve_entity\`** per account: cap to N calls per minute with jitter.
2. **Respect \`flood_wait_until\`** at the lease level: \`acquire_by_phone\` already checks \`flood_wait_until > now\`, but \`collect_channel_resolve_username\` should abort early and reschedule the whole collect cycle instead of hammering remaining accounts.
3. **Fallback to cached entity** in \`channels\` table where \`access_hash\` is present — avoid resolve entirely for known channels.
4. **Back off globally** once a single account hits FloodWaitError > 300s: stop dispatching resolve to *other* accounts for at least \`min(wait_seconds, 3600)\` to avoid poisoning the whole pool.

## Related

Follow-up of #463 (pipeline result observability). The observability fix now surfaces \`no_available_client\` node-errors, but root cause (pool-wide flood storm) needs addressing here.

axisrow:
## Summary
- add process-local backoff for long `collect_channel_resolve_username` FloodWaits
- keep existing `pool.report_flood()` behavior while preventing immediate rotation across the account pool
- leave short FloodWaits on the existing rotation path
- cover production-like `TelegramTransportSession` behavior and the 3600s backoff cap

Closes #464

## Verification
- `pytest tests/test_collector.py -q`
- `pytest tests/test_account_lease_pool.py tests/test_flood_wait.py tests/test_backends.py -q`

### DIFF
--- a/src/telegram/collector.py
+++ b/src/telegram/collector.py
@@ -4,7 +4,7 @@ import asyncio
 import json
 import logging
 from collections.abc import Awaitable, Callable
-from datetime import date, datetime, timezone
+from datetime import date, datetime, timedelta, timezone
 
 from telethon.errors import FloodWaitError, UsernameInvalidError, UsernameNotOccupiedError
 from telethon.tl.types import (
@@ -43,6 +43,10 @@ from src.telegram.flood_wait import HandledFloodWaitError, run_with_flood_wait
 from src.telegram.notifier import Notifier
 
 logger = logging.getLogger(__name__)
+
+RESOLVE_USERNAME_OPERATION = "collect_channel_resolve_username"
+RESOLVE_USERNAME_GLOBAL_BACKOFF_THRESHOLD_SEC = 300
+RESOLVE_USERNAME_GLOBAL_BACKOFF_MAX_SEC = 3600
 
 
 class NoActiveStatsClientsError(RuntimeError):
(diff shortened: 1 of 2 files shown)


axisrow:
@codex review. Focus on critical issues: bugs, security vulnerabilities, logical errors, data loss risks, performance problems. Do NOT nitpick style, naming conventions, minor formatting, or subjective preferences — only flag issues that could break functionality or cause real harm in production.

PR #518, opened 2026-04-29
fix(agent): route zai through openai compatible endpoint

axisrow:
## Summary
- switch Z.AI deepagents runtime from Anthropic-compatible routing to the official OpenAI-compatible GLM endpoint
- normalize saved legacy Z.AI Anthropic base URLs to the current OpenAI-compatible base URL
- update provider registration and regression tests for the Z.AI endpoint path

## Tests
- ruff check src/ tests/ conftest.py
- pytest tests/test_agent.py -v -k "zai or deepagents"
- pytest tests/test_agent_provider_service.py -v -k "zai or canonical_endpoint_fingerprint or normalize_plain_fields"
- pytest tests/test_provider_service_adapters.py -v -k zai
- pytest tests/test_agent.py tests/test_agent_provider_service.py tests/test_provider_service.py tests/test_provider_service_adapters.py tests/test_provider_adapters.py -v

Closes #516

PR #519, opened 2026-04-29
fix(agent): use zai coding endpoint

### LINKED ISSUE #516: fix(web): agent chat unavailable in web process
## Problem

The web UI shows this error when using Agent chat:

> Чат недоступен в web-процессе. Agent chat is only available in the worker process. Run `python -m src.main worker` alongside the web server to enable it.

This blocks Agent chat from the web dashboard unless a separate worker process is running. The current single-command development/runtime flow expects `python -m src.main serve` to provide the web UI plus embedded worker by default, so this looks like a regression or incorrect runtime-mode detection.

## Expected Behavior

Agent chat should work from the web UI when the app is started with the normal embedded-worker server command:

```bash
python -m src.main serve
```

If the app is intentionally started with `--no-worker`, the message should remain accurate and actionable.

## Actual Behavior

The web UI reports that chat is unavailable in the web process and instructs the user to run a separate worker process.

## Suggested Investigation

- Check how the web process determines whether Agent chat is available.
- Verify embedded-worker mode is detected correctly.
- Confirm `python -m src.main serve` initializes whatever Agent chat dependency/state the route checks.
- Keep split deployment behavior unchanged for `serve --no-worker` plus separate `worker`.

## Acceptance Criteria

- Agent chat works in the dashboard under normal `python -m src.main serve`.
- `serve --no-worker` still returns an appropriate worker-only message if no worker is available.
- Add or update route/runtime tests for both embedded-worker and web-only modes.

axisrow:
## Summary
- switch Z.AI default runtime base URL to the Coding Plan OpenAI-compatible endpoint
- keep general Z.AI endpoint as an explicit constant for future/custom use
- fetch Z.AI model catalog from the same coding endpoint used by runtime
- update legacy URL normalization expectations

## Verification
- Official Z.AI docs checked: General API uses https://api.z.ai/api/paas/v4, Coding Plan tools use https://api.z.ai/api/coding/paas/v4
- Live smoke with local ZAI_API_KEY:
  - https://api.z.ai/api/paas/v4/chat/completions -> 429 insufficient balance/resource package
  - https://api.z.ai/api/coding/paas/v4/chat/completions -> 200 with glm-5-turbo
  - updated provider adapter returned ok with glm-5-turbo and glm-4.7

## Tests
- ruff check src/ tests/ conftest.py
- pytest tests/test_agent.py -v -k "zai or deepagents"
- pytest tests/test_agent_provider_service.py -v -k "zai or canonical_endpoint_fingerprint or normalize_plain_fields"
- pytest tests/test_provider_service_adapters.py -v -k zai

Closes #516

### DIFF
--- a/src/agent/provider_registry.py
+++ b/src/agent/provider_registry.py
@@ -4,7 +4,8 @@ from dataclasses import dataclass, field
 
 from src.agent.models import CLAUDE_MODELS
 
-ZAI_DEFAULT_BASE_URL = "https://api.z.ai/api/paas/v4"
+ZAI_DEFAULT_BASE_URL = "https://api.z.ai/api/coding/paas/v4"
+ZAI_GENERAL_BASE_URL = "https://api.z.ai/api/paas/v4"
 ZAI_LEGACY_ANTHROPIC_BASE_URLS = {
     "https://api.z.ai/api/anthropic",
     "https://api.z.ai/api/anthropic/v1",
--- a/src/services/agent_provider_service.py
+++ b/src/services/agent_provider_service.py
@@ -805,7 +805,7 @@ class AgentProviderService:
         # Use the native Z.AI API endpoint with Bearer auth instead.
         headers = {"Authorization": f"Bearer {api_key}"}
         payload = await self._fetch_json(
-            "https://api.z.ai/api/paas/v4/models", headers=headers
+            f"{ZAI_DEFAULT_BASE_URL}/models", headers=headers
         )
         return [
             str(item.get("id", "")).strip()
--- a/tests/test_agent_provider_service.py
+++ b/tests/test_agent_provider_service.py
@@ -910,7 +910,7 @@ async def test_fetch_live_models_zai_uses_bearer_auth(db, monkeypatch):
     models = await service._fetch_live_models(spec, cfg)
 
     assert models == ["glm-5"]
-    assert captured["url"] == "https://api.z.ai/api/paas/v4/models"
+    assert captured["url"] == "https://api.z.ai/api/coding/paas/v4/models"
     assert captured["headers"] == {"Authorization": "Bearer zai-key"}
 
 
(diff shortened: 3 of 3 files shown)


axisrow:
@claude review. Focus on critical issues only: endpoint mapping bugs, provider compatibility, regressions, security vulnerabilities, data loss risks. Context: Z.AI Coding Plan uses https://api.z.ai/api/coding/paas/v4 for OpenAI-compatible tools; general /paas/v4 returns 429 for Coding Plan quota.

axisrow:

axisrow:
Thanks for catching the migration concern, but flagging this as CONFLICTING with the policy decided in the PR stack:

- This PR (#519) is the narrow endpoint constant fix from issue #516.
- It is the base of a stack: #519 ← #522 (test coverage, issue #520) ← #524 (issue #523, endpoint policy).
- Issue #523 explicitly decided that the **default** Z.AI base URL stays on the **general** endpoint (`https://api.z.ai/api/paas/v4`) and **Coding Plan is opt-in only**. That is exactly what #524 implements, including `ZAI_DEFAULT_BASE_URL = ZAI_GENERAL_BASE_URL` and validation errors for legacy Anthropic URLs.

Adding `ZAI_GENERAL_BASE_URL` to `ZAI_LEGACY_ANTHROPIC_BASE_URLS` here would:
1. Silently route every non-Coding-Plan user to a Coding-only endpoint, which Z.AI docs explicitly recommend against ("For other use cases, we recommend using the general API endpoint").
2. Be reverted by #524 within the same stack, producing churn rather than safety.

Existing users with empty `base_url` (the most common case after #518) get the new default through `normalize_zai_base_url` correctly. Users with `paas/v4` explicitly stored will continue using the general endpoint after the stack lands — which is the documented recommended path, not a 429 trap.

No fix applied here; concern is addressed by #524 in the same review cycle.

PR #526, opened 2026-04-30
fix(agent): default Z.AI to general endpoint

### LINKED ISSUE #523: chore(agent): revisit normalize_zai_base_url and ZAI_DEFAULT_BASE_URL choice
## Problem

Current Z.AI integration has two design choices that conflict with official Z.AI documentation:

### 1. Coding-Plan endpoint is hardcoded as default for all Z.AI users

`src/agent/provider_registry.py:7` sets `ZAI_DEFAULT_BASE_URL = "https://api.z.ai/api/coding/paas/v4"`. This is the GLM Coding Plan endpoint.

Per Z.AI docs (https://docs.z.ai/api-reference):

> The GLM Coding Plan API endpoint is intended for supported tools only. **For other use cases, we recommend using the general API endpoint.**

The general endpoint is `https://api.z.ai/api/paas/v4`, which we already define as `ZAI_GENERAL_BASE_URL` (`provider_registry.py:8`) but never default to. Users who don't have a Coding Plan subscription will silently get routed to a Coding-only endpoint, which is at best confusing and at worst rejects their requests.

### 2. `normalize_zai_base_url` silently rewrites Anthropic-format URL to OpenAI-format

`src/agent/provider_registry.py:15-19`:

```python
def normalize_zai_base_url(base_url: str = "") -> str:
    normalized = (base_url or "").strip().rstrip("/")
    if not normalized or normalized in ZAI_LEGACY_ANTHROPIC_BASE_URLS:
        return ZAI_DEFAULT_BASE_URL
    return normalized
```

`https://api.z.ai/api/anthropic` is documented (https://docs.z.ai/scenario-example/develop-tools/claude) as an **Anthropic-compatible proxy** (`ANTHROPIC_BASE_URL` for Claude Code) — it expects Anthropic message format, not OpenAI `/chat/completions`. Rewriting it to a Coding-Plan OpenAI endpoint mixes two incompatible API formats. A user who configured an Anthropic URL by mistake gets a silently-mutated config and an opaque downstream behavior change instead of a clear validation error.

## Investigation findings

`normalize_zai_base_url` is currently called from several places:

- `src/agent/manager.py:1211` — `DeepagentsBackend._build_agent` (runtime adapter construction)
- `src/services/agent_provider_service.py:668-671` — `canonical_endpoint_fingerprint`
- `src/services/agent_provider_service.py:895-896` — `_normalize_plain_fields` (persists normalized URL on save)
- `src/services/provider_service.py:268-271` — `_build_adapter_for_config`

`ZAI_DEFAULT_BASE_URL` is also used by:

- `src/services/agent_provider_service.py:803-810` — `_fetch_zai_models` (hardcodes `{ZAI_DEFAULT_BASE_URL}/models`)
- `src/services/provider_service.py:76-78` — env-var-based provider registration via `ZAI_API_KEY`
- `src/agent/provider_registry.py:280` — UI placeholder for the Z.AI Base URL field

So the change touches: provider registration, persistence, runtime adapter, model refresh, and the settings UI.

## Required Work

Decide and implement a coherent default-base-url and normalization policy. Three suggested decisions:

1. **Default base URL: switch from Coding to General.** Set `ZAI_DEFAULT_BASE_URL = "https://api.z.ai/api/paas/v4"`. Add a separate `ZAI_CODING_BASE_URL` constant (or leave `ZAI_GENERAL_BASE_URL` and rename appropriately). Coding endpoint becomes opt-in via explicit user configuration, not silent default. Update the UI placeholder accordingly.

2. **Stop silently rewriting Anthropic URLs.** When a user configures `https://api.z.ai/api/anthropic[/v1]` for the Z.AI provider, surface a validation error like "This URL is the Anthropic-compatible proxy. Configure the `anthropic` provider with this URL instead, or use the OpenAI-compatible endpoint https://api.z.ai/api/paas/v4." Do not mutate the field on save. (`canonical_endpoint_fingerprint` and `_normalize_plain_fields` need to stop calling `normalize_zai_base_url`, and validation should reject the legacy URL with a helpful message.)

3. **`_fetch_zai_models` should follow the user's base_url, not hardcode the default.** Today it hits `{ZAI_DEFAULT_BASE_URL}/models` regardless of what the user configured. If we make Coding opt-in, `/models` must be fetched from whatever base_url the user actually picked. (Otherwise we'd refresh against one endpoint and chat against another — the same kind of mismatch that triggered #520.)

## Acceptance Criteria

- A fresh Z.AI provider config with empty `base_url` defaults to **the general endpoint**, not the Coding endpoint, on both runtime and `/models` paths.
- Configuring `https://api.z.ai/api/anthropic[/v1]` produces a **validation error** with a clear message; the value is not silently rewritten in DB.
- Users on the Coding Plan can opt in by explicitly setting `https://api.z.ai/api/coding/paas/v4` in the UI; runtime and `/models` both honor that exact URL.
- Existing tests at `tests/test_provider_runtime_integration.py`, `tests/test_agent_provider_service.py`, `tests/test_cli_provider.py`, `tests/test_agent.py` are updated to reflect the new defaults and validation behavior.
- Adds a regression test that fails if `_fetch_zai_models` hardcodes the default URL again instead of honoring the configured one.
- Documentation note added (CLAUDE.md or short Z.AI README section) pointing at the official docs and explaining when to pick which endpoint.

## Out of scope

- Rewriting how other providers normalize base URLs.
- Changing how `anthropic` provider is configured (the `anthropic` provider in `PROVIDER_SPECS` is correct as-is — just point users there from the validation error).

## References

- Z.AI API reference: https://docs.z.ai/api-reference
- Z.AI Claude Code integration guide: https://docs.z.ai/scenario-example/develop-tools/claude
- Predecessor issues: #516 (runtime fix), #518 (route through openai-compat), #520 (test coverage)

axisrow:
Closes #523. Reopened from #524 after the stack base merged via squash (which auto-closed the original PR).

## Summary
- default Z.AI OpenAI-compatible base URL to the general endpoint
- keep Coding Plan endpoint as explicit opt-in
- reject Z.AI Anthropic-compatible proxy URLs for the `zai` provider with a clear validation error
- make Z.AI `/models` refresh use the configured base URL
- document Z.AI endpoint selection in provider docs

## Validation
- `pytest tests/test_provider_runtime_integration.py -v`
- `pytest tests/test_agent_provider_service.py -v -k zai`
- `pytest tests/test_agent.py -v -k zai`
- `pytest tests/test_cli_provider.py -v`
- `pytest tests/test_cli_agent_commands.py -v`
- `pytest tests/test_provider_service.py tests/test_provider_service_adapters.py tests/test_agent_provider_service.py -v`

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -12,6 +12,7 @@ Supported env vars (examples)
 - FIREWORKS_BASE, FIREWORKS_API_KEY — Fireworks inference endpoint
 - DEEPSEEK_BASE, DEEPSEEK_API_KEY — DeepSeek endpoint
 - TOGETHER_BASE, TOGETHER_API_KEY — Together API base
+- ZAI_API_KEY — Z.AI API key (OpenAI-compatible chat completions)
 
 Env-based registration
 
@@ -22,6 +23,13 @@ Settings UI
 - Use the Settings → Developer → Providers form to add, edit or remove provider configurations. Secret fields are encrypted before storage using SESSION_ENCRYPTION_KEY; when this key is not set the UI is read-only for provider secrets.
 - The provider registry uses canonical provider names (e.g. "openai", "cohere", "ollama", "huggingface", "context7").
 - For local Ollama testing, set OLLAMA_BASE to your server URL and optionally OLLAMA_DEFAULT_MODEL.
+
+Z.AI endpoints
+
+- The `zai` provider uses the OpenAI-compatible Z.AI API. Empty Base URL defaults to the general endpoint: `https://api.z.ai/api/paas/v4`.
+- GLM Coding Plan users can opt in by setting Base URL to `https://api.z.ai/api/coding/paas/v4`; model refresh and chat runtime both use the configured endpoint.
+- Do not use `https://api.z.ai/api/anthropic` or `/anthropic/v1` with the `zai` provider. That URL is the Anthropic-compatible proxy for tools such as Claude Code; configure the `anthropic` provider for that endpoint instead.
+- References: https://docs.z.ai/api-reference/introduction and https://docs.z.ai/devpack/tool/claude
 
 Context7 & operator caution
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ dependencies = [
     "langchain==1.2.12",
     "langchain-core==1.2.19",
     "langchain-ollama==1.0.1",
+    "langchain-openai==1.1.11",
     "pytest==9.0.2",
     "pytest-timeout==2.4.0",
     "pytest-xdist==3.8.0",
--- a/src/agent/manager.py
+++ b/src/agent/manager.py
@@ -39,7 +39,13 @@ from src.agent.prompt_template import (
     build_prompt_template_context,
     render_prompt_template,
 )
-from src.agent.provider_registry import ProviderRuntimeConfig, normalize_zai_base_url
+from src.agent.provider_registry import (
+    ZAI_CODING_BASE_URL,
+    ZAI_DEFAULT_BASE_URL,
+    ProviderRuntimeConfig,
+    is_zai_legacy_anthropic_base_url,
+    normalize_zai_base_url,
+)
 from src.config import AppConfig
 from src.database import Database
 from src.services.agent_provider_service import (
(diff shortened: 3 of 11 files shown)


axisrow:

axisrow:
Triage:

**1 (empty base_url migration)** — SKIP. Reviewer's own framing: "This is the stated intent of the PR". Issue #523 explicitly decided that Z.AI default switches to the general endpoint and Coding Plan becomes opt-in. The behavior change is the feature, not a regression.

**2 (legacy URL passthrough)** — SKIP. By design after #523. All four call sites (`_build_agent`, `_build_adapter_for_config`, `validate_provider_config`, `_fetch_zai_models`) guard with `is_zai_legacy_anthropic_base_url` before calling `normalize_zai_base_url`, so the contract change is safe in practice. Reviewer: "low risk given all callers are accounted for".

**3 (redundant `rstrip('/')`)** — SKIP. Reviewer's own framing: "harmless but dead code. Minor." Brief excludes minor cosmetics.

**4 (validate/build duplication)** — Reviewer: "No issue here". This is intentional defense-in-depth.

Reviewer verdict: "Ready to merge — no blocking issues". No FIX verdicts. Proceeding to CI.

PR #612, opened 2026-05-27
fix(web): repair search translations and post stats

### LINKED ISSUE #603: Ошибки в отделе "Поиск"
https://tg-content-factory.46.225.99.67.sslip.io/search?q=Paris&channel_id=&date_from=&date_to=&mode=local
Отдел: Поиск
Шаги: Запрос - Paris
1. В конце поста кнопка перевода не работает

До`
Image

После`
Image

2. Статистика поста не обновляется 

Image

Image

axisrow:
## Summary
- Fix live search results so translated posts have DB-backed message IDs and usable translate buttons.
- Refresh mutable post engagement fields on duplicate message saves so search stats stay current.
- Add regression coverage for translation rendering, Telegram search metadata, and duplicate stats refresh.

## Test plan
- `pytest tests/test_transformers.py tests/test_search.py tests/routes/test_search_routes.py tests/repositories/test_messages_repository.py -v`

Fixes #603

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/database/repositories/messages.py
+++ b/src/database/repositories/messages.py
@@ -223,6 +223,14 @@ class MessagesRepository:
         ]
         try:
             existing_keys = await self._existing_message_keys(messages)
+        except Exception as exc:
+            logger.warning(
+                "Could not pre-fetch existing message keys for refresh, proceeding without: %s",
+                exc,
+            )
+            existing_keys = set()
+
+        try:
             cur = await self._database.executemany_write(
                 """INSERT OR IGNORE INTO messages
                    (channel_id, message_id, sender_id, sender_name,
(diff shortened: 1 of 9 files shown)


axisrow:

axisrow:
Triage of review feedback:

**Issue 1 — unconditional `_refresh_existing_messages` in batch path** → FIXED. The batch path now pre-fetches the existing `(channel_id, message_id)` keys and only refreshes that subset, mirroring the single-message path semantics. Pre-SELECT is chunked at 400 pairs to stay under SQLite's `SQLITE_LIMIT_VARIABLE_NUMBER`.

**Issue 2 — SQLite variable limit in `get_messages_by_channel_message_ids`** → SKIP for this PR. Confirmed the latent footgun exists, but no realistic caller reaches it: Telegram search returns ≤100/page, the agent tool defaults to limit=50, and the only caller is `_load_persisted_messages` which always passes the just-fetched batch. Worth tracking as a separate hardening task; out of scope for the search/translate fix this PR is targeting.

**Issue 3 — silent partial drop in `_load_persisted_messages`** → FIXED. Replaced `return persisted or messages` with a key-based merge that returns the persisted row when available and falls back to the original message for any missing key, preserving input order. Added `try/except` around the DB call (logs and returns the originals on error) and a `logger.warning` when persistence returns a partial result.

**Issue 4 — synchronous `detect_language` per message** → SKIP for this PR. `langdetect.detect()` is pure-Python n-gram analysis on ≥8-char text (≈1–5 ms/call); for typical search batches (≤50 messages) total cost is dominated by Telegram RTT. The collector path (`src/telegram/collector.py:905`) already calls it synchronously, so changing only the search path would be inconsistent. If we want CPU-bound work off the loop, it should be done in one place (a `to_thread` wrapper at the transformer call site), tracked separately.

axisrow:

axisrow:

PR #661, opened 2026-05-31
fix(async): HTTP timeouts and non-blocking file writes in adapters (#633)

axisrow:
## Summary

Next batch of fixes from the #633 bug report — hardening of outbound HTTP and file I/O in the provider/image adapters and the photo-loader upload path. Addresses bugs **#10, #11, #12, #25**.

## Changes

- **`provider_adapters.py` (#10)** — every `aiohttp.ClientSession` now uses a shared `_HTTP_TIMEOUT` (`ClientTimeout(total=120)`). Previously sessions had no timeout, so a stalled LLM/image upstream could hang the calling coroutine indefinitely (blocking collection queue, web requests, scheduler).
- **HuggingFace image adapter (#11, #12)** — create the output directory before writing (`out.mkdir(parents=True, exist_ok=True)`) to avoid `FileNotFoundError` when `data/images` is absent, and write the image via `asyncio.to_thread(...)` so the blocking `write_bytes` no longer stalls the event loop.
- **`photo_loader/forms.py::persist_uploads` (#25)** — write uploaded bytes via `asyncio.to_thread(...)` instead of a synchronous `write_bytes` in an async context.

## Out of scope

- **#8** (double-consume of response body) — already fixed in #639; current code reads `resp.text()` only on the error branch.
- **#3** (react_agent blocking `urlopen`) — `OllamaReActAgent` is synchronous and invoked from the deepagents backend; the blocking-I/O concern depends on the caller's threading and is left for a separate change.


axisrow:
@codex review. Focus on critical issues: bugs, security vulnerabilities, logical errors, data loss risks, performance problems. Do NOT nitpick style, naming conventions, minor formatting, or subjective preferences — only flag issues that could break functionality or cause real harm in production.

PR #748, opened 2026-06-03
test(747): verify-before-delete guard for live CLI test cleanup

### LINKED ISSUE #747: Live CLI test cleanup could delete content it did not create (verify-before-delete)
## Summary

The cleanup logic of the live (real Telegram) CLI integration tests in
`tests/cli_real_tg_integration/` could delete or overwrite content that the test
did **not** itself create. A cleanup routine captured an object id and then
deleted it without first confirming that the object actually belonged to the
test run. A loosely-parsed id pointing at the wrong message therefore led to a
**non-test message being deleted** during cleanup.

This is a data-loss safety bug, not a functional one — the product behaved
correctly; the test harness lacked a guarantee that destructive cleanup only
ever touches what the test created.

## Root cause

- Cleanup deleted/edited by an id captured from CLI output (or read from the DB)
  **without verifying ownership** of the target before acting.
- Telegram's server-side protection only blocks deleting *other people's*
  messages; it does **not** prevent deleting a *different one of your own*
  messages when the id is wrong.
- One cleanup path (`scheduler clear-pending`) was **global** — it cancelled
  *all* pending tasks in the queue, not only the ones the test enqueued.
- No unique per-run marker was checked before deletion, so a wrong/ambiguous id
  could not be distinguished from the intended target.

## Fix (verify-before-delete)

Every destructive cleanup now proves ownership before acting:

1. **Unique nonce** (`uuid4().hex`) embedded in the text/caption of every
   message a test creates.
2. **Re-read before delete**: cleanup reads the target back via
   `messages read --live` and deletes **only** when the message id matches *and*
   the text carries the test's own nonce. Any read error, FLOOD_WAIT, missing
   message, or nonce mismatch → **do not delete** (fail-safe leak instead of a
   risky deletion).
3. **Targeted scheduler cleanup**: replaced the global `clear-pending` with a
   before/after snapshot diff that cancels **only** the task ids this test
   created.
4. **Channel cleanup** (create → leave): the channel title is made unique and
   verified before the channel is left/deleted.
5. New shared helpers in `conftest.py`: `make_cli_nonce`,
   `cli_verify_message_nonce`, `assert_safe_to_delete` (returns the verified
   subset of ids; never deletes itself), `cli_verify_channel_title`, and
   targeted scheduler-cancel helpers.

The guarantee is now **structural**: cleanup can only ever remove content that
provably originated from the same test run.

## Verification

- Offline fail-safe proof: a message without the nonce, a wrong/absent id, and a
  tampered nonce are each rejected and never authorized for deletion.
- `ruff`, the real-telegram policy suite, and SQL-convention checks pass.
- Live re-run of the affected categories confirms verify-then-delete works on
  freshly created content.

## Scope / notes

- Reversal-only cleanups (unpin / unarchive / clear-reaction) operate on a
  verified fixture id and cannot delete content, so they are unchanged.
- Pipeline-published messages cannot carry a test nonce (text is generated by the
  pipeline); cleanup there verifies the published message against the run's own
  generated text before deleting.

axisrow:
## Summary

Live CLI test cleanup (`tests/cli_real_tg_integration/`) could delete or
overwrite content it did **not** create: cleanup captured an id and deleted it
without confirming ownership of the target, and `scheduler clear-pending`
cancelled *all* pending tasks globally. A wrong/ambiguous id could therefore
remove a non-test message. See #747.

## What changed

Destructive cleanup now proves ownership before acting:

- **Nonce per run**: every test-created message embeds a unique `uuid4().hex` in
  its text/caption.
- **Re-read before delete**: cleanup reads the target back via
  `messages read --live` and deletes only when the message id matches **and** the
  text carries the test's own nonce. Any read error, FLOOD_WAIT, missing message,
  or nonce mismatch → **do not delete** (fail-safe leak, never a risky deletion).
- **Targeted scheduler cleanup**: global `scheduler clear-pending` replaced with a
  before/after snapshot diff that cancels only the task ids this test enqueued.
- **Channel cleanup**: a unique channel title is verified before the test-created
  channel is left/deleted.
- **Pipeline-published cleanup**: verified against the run's own generated text
  (a test nonce cannot be injected into pipeline-generated content).

New `conftest.py` helpers: `make_cli_nonce`, `cli_verify_message_nonce`,
`assert_safe_to_delete` (returns the verified id subset; never deletes itself),
`cli_verify_channel_title`, `fetch_pipeline_run_text`,
`snapshot_pending_collection_task_ids`, `cancel_collection_tasks`.

### Policy contract

Verification reads and targeted `scheduler task-cancel` calls go through a raw
`_capture_cli` helper named **outside** the policy auditor's
`run_cli` / `run_cli_popen` / `cli_run_direct` set, so they are not treated as
blind-delete cleanup calls. The blind-delete literals
(`dialogs delete-message`, `dialogs leave`, `photo-loader batch-cancel`) remain
in each test's `finally` as `cli_run_direct(...)` with a producer in the same
test, as `tests/test_real_telegram_policy.py` requires.

## Verification

- `ruff check` clean; `pytest tests/test_real_telegram_policy.py` (41) and
  `tests/test_sql_conventions.py` (6) pass; `--collect-only` collects 110 tests.
- Offline fail-safe proof: a message without the nonce, a wrong/absent id, and a
  tampered nonce are each rejected and never authorized for deletion.
- Live re-run of the affected categories validates verify-then-delete on freshly
  created content (sequential, shared session strings).

## Scope

Reversal-only cleanups (unpin / unarchive / clear-reaction) act on a verified
fixture id and cannot delete content, so they are unchanged.

Closes #747

### DIFF
--- a/tests/cli_real_tg_integration/conftest.py
+++ b/tests/cli_real_tg_integration/conftest.py
@@ -530,8 +530,16 @@ class VerifyResult:
 def _parse_live_message_blocks(stdout: str) -> dict[int, str]:
     """Map message_id -> joined text body from `messages read --live` output.
 
-    The live format is a header line ``[<date>] #<id> <sender>`` followed by
-    indented (two-space) body/text lines until the next header or a blank line.
+    The live format is a header line ``[<date>] #<id> <sender>`` followed by the
+    message body. `_print_live_messages` (src/cli/commands/messages.py) prints the
+    whole body with a single ``print(f"  {text[:500]}")``, so ONLY the first
+    physical line of a multi-line body carries the two-space indent — continuation
+    lines (after embedded newlines) are unindented, and blank lines may appear
+    between paragraphs. We therefore accumulate every line after a header until the
+    NEXT header line, stripping the optional two-space prefix. Extra trailing lines
+    (the inter-message blank, an optional ``reaction users:`` line) are harmless:
+    the caller only substring-checks the body for its nonce/marker, which those
+    lines never contain.
     """
     blocks: dict[int, list[str]] = {}
     current: int | None = None
@@ -543,14 +551,7 @@ def _parse_live_message_blocks(stdout: str) -> dict[int, str]:
             continue
         if current is None:
             continue
-        if raw_line.startswith("  "):
-            blocks[current].append(raw_line[2:])
-        elif raw_line.strip() == "":
-            # A blank line is part of a multi-paragraph body, not a block
-            # terminator: keep accumulating so a nonce in a later paragraph is
-            # still found. Only a new header line (handled above) starts a new
-            # block. Trailing blank lines are stripped by the join+strip below.
-            blocks[current].append("")
+        blocks[current].append(raw_line[2:] if raw_line.startswith("  ") else raw_line)
     return {mid: "\n".join(lines).strip("\n") for mid, lines in blocks.items()}
 
 
--- a/tests/cli_real_tg_integration/conftest.py
+++ b/tests/cli_real_tg_integration/conftest.py
@@ -546,8 +546,12 @@ def _parse_live_message_blocks(stdout: str) -> dict[int, str]:
         if raw_line.startswith("  "):
             blocks[current].append(raw_line[2:])
         elif raw_line.strip() == "":
-            current = None
-    return {mid: "\n".join(lines) for mid, lines in blocks.items()}
+            # A blank line is part of a multi-paragraph body, not a block
+            # terminator: keep accumulating so a nonce in a later paragraph is
+            # still found. Only a new header line (handled above) starts a new
+            # block. Trailing blank lines are stripped by the join+strip below.
+            blocks[current].append("")
+    return {mid: "\n".join(lines).strip("\n") for mid, lines in blocks.items()}
 
 
 def cli_verify_message_nonce(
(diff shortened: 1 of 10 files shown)


axisrow:

axisrow:
Thanks — all findings addressed in 216ff04:

- **Snapshot failure → empty baseline (Claude #1 / Codex P1 conftest:733):** `snapshot_pending_collection_task_ids` now returns `(ids, ok)`. A failed read no longer falls back to `set()`; both scheduler/collect cleanups skip cancellation and leak a warning when either snapshot is unreliable, so `after - {}` can never cancel every pending task.
- **Diff proves "new", not "mine" (Codex P1 collect:42):** acknowledged. There is no per-test ownership column on `collection_tasks`, and adding one is a product change out of scope here. The diff is exact in practice because this suite runs **sequentially** (shared session strings) and the live CLI tests run **no worker**, so within the test window the test's own `collect`/`trigger` is the only enqueuer. I've documented that assumption in both tests and combined it with the abort-on-unreliable-snapshot guard above as the strongest in-scope mitigation.
- **Pipeline marker not unique (Claude #2):** `distinctive_text_fragment` now requires a ≥40-char line (max 80) and returns `None` (cleanup skipped) when none qualifies, making cross-run collisions negligible.
- **Blank line truncates body (Claude #3):** `_parse_live_message_blocks` no longer closes a block on blank lines, so a nonce in a later paragraph of a multi-paragraph body is still found.

Verified offline (ruff, policy suite 41, sql conventions, collect-only 110) plus an extended fail-safe proof covering multi-paragraph nonce, snapshot-failure→skip, and the raised fragment thresholds.

@claude @codex review.

axisrow:
Fixed the P2 in e0495b2.

**Codex P2 (conftest.py:548 — unindented continuation lines):** correct and confirmed against `_print_live_messages` — it emits the whole body with a single `print(f"  {text[:500]}")`, so only the first physical line is indented; continuation lines after embedded newlines are unindented. The parser now accumulates every line after a header until the next header (stripping the optional two-space prefix), so a nonce/marker on any later line is found. Trailing inter-message blanks and an optional `reaction users:` line are harmless — the body is only substring-checked for the nonce.

Offline proof added for the exact shape (nonce on an unindented later line is now found) while ownership invariants still hold fail-safe (wrong id, tampered nonce, and another account's message are all rejected). Also refreshed the stale docstring (Claude's nit).

@claude @codex review.

axisrow:
@codex review — please review the latest commit e0495b2 (the unindented-continuation-line parser fix). Earlier P1/P2 findings are all addressed.

PR #758, opened 2026-06-03
fix(codex): require writable Codex runtime in codex_available()

axisrow:
## Summary

`codex_available()` previously checked only that the Codex SDK was installed and `~/.codex/auth.json` existed. In environments where the Codex runtime is **not writable** (sandboxed CI, restricted filesystem), this returned `True` — the provider registered as usable, but every live agent turn through the Codex backend **stalled until `agent.total_timeout` (300s)**: Codex spawns the `mcp-server` subprocess and calls project tools (verified via `CallToolRequest` in logs), but can't update its runtime state DB, so the turn never completes and gets killed by the timeout.

This made `test_agent_test_tools_via_codex_backend` (opt-in `RUN_CODEX_CLI_LIVE`, never run in CI) fail by timeout (~12 min) on any machine with a read-only Codex home.

## Changes

- `codex_available()` now also verifies the Codex runtime is writable:
  - `$CODEX_HOME` (or `~/.codex`) directory must be writable
  - any existing `state*.sqlite*` files must be writable
- `$CODEX_HOME` is now honored as authoritative (was hardcoded to `~/.codex`).
- New helpers: `_codex_home_path()`, `_codex_path_writable()`, `_codex_runtime_state_writable()`.

When the runtime isn't usable the provider stays unregistered, and `_require_codex()` in the live suite skips the cases cleanly instead of hanging.

## Tests

4 new unit tests covering: writable home (True), missing auth (False), read-only home (False), read-only state DB (False). Full image/codex/provider suites green (150 passed).

### DIFF
--- a/src/services/provider_adapters.py
+++ b/src/services/provider_adapters.py
@@ -422,7 +422,7 @@ def _codex_runtime_state_writable(codex_home: Path) -> bool:
         if not codex_home.is_dir() or not _codex_path_writable(codex_home):
             return False
         for path in codex_home.glob("state*.sqlite*"):
-            if path.exists() and not _codex_path_writable(path):
+            if not _codex_path_writable(path):
                 return False
     except OSError:
         return False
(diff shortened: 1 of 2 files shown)


axisrow:

PR #763, opened 2026-06-04
fix(agent): SSE keepalive + shared total deadline instead of idle timeout

axisrow:
## Summary

The agent chat SSE wrapper previously aborted the stream after 180s without a chunk from the backend (`_SSE_IDLE_TIMEOUT`), which killed long-but-healthy agent turns. This replaces the per-chunk idle timeout with:

- a **15s keepalive cadence** (`_SSE_KEEPALIVE_INTERVAL`) that emits a visible `{"type": "status", "text": "Агент всё ещё работает... Nс"}` frame instead of cancelling, and
- a **hard total deadline** read from the shared `config.agent.total_timeout` (default 300s) — the same knob the `ClaudeSdkBackend`/`codex_backend` already enforce, so the web layer no longer carries an uncoordinated magic constant.

## Quality cleanups (from `/simplify`)

- Reuse one `pending_chunk_task` across keepalive wakeups instead of cancel-and-recreate, so non-resumable backend work is not abandoned.
- Read `config.agent.total_timeout` directly (typed Pydantic `int`) instead of a defensive `getattr`+`float()` chain with a duplicate default constant — matching `manager.py:642` / `codex_backend.py:164`.
- Emit the keepalive frame via `safe_json_dumps`, consistent with the rest of the agent SSE stack.
- Drop the throwaway temp-var when reading the chunk result.

## Test plan

- `pytest tests/routes/test_agent_routes_streaming.py tests/test_cli_agent_commands.py` — 55 passed.
- New coverage: keepalive does not cancel an active slow stream; a second message on the same thread survives keepalive after the first turn; total-timeout cancels with bounded wait; CLI consumes `status`/`countdown` frames on a second turn.
- `ruff check` clean.

### DIFF
--- a/src/web/agent/handlers.py
+++ b/src/web/agent/handlers.py
@@ -426,6 +426,15 @@ async def chat(request: Request, thread_id: int):
                 if done_data is not None:
                     break
         finally:
+            if pending_chunk_task is not None and not pending_chunk_task.done():
+                pending_chunk_task.cancel()
+                pending_chunk_task.add_done_callback(_consume_abandoned_next_chunk)
+                done, _ = await asyncio.wait({pending_chunk_task}, timeout=1.0)
+                if not done:
+                    logger.debug(
+                        "Agent stream next-chunk task did not finish after cancellation for thread %d",
+                        thread_id,
+                    )
             if pending_chunk_task is None or pending_chunk_task.done():
                 try:
                     await agen.aclose()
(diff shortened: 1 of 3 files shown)


axisrow:

axisrow:
@claude @codex review\n\nSecond review round after addressing Claude's finding: the SSE handler now cancels and drains a pending next-chunk task on client disconnect before closing the async generator, with a focused regression test for disconnect cleanup. Full local validation passed before push: ruff, route/CLI focused tests, parallel pytest lane, serial aiosqlite lane.

PR #858, opened 2026-06-15
fix(parsing): chat filter bare-id, invite links, channel deactivation (835#7,#17,#8)

axisrow:
## Что
- **835#7** `_normalize_token`: `t.me/c/<id>` → bare-positive id (совпадает с `channel_id` в БД).
- **835#17** `parsers`: убран `\+?` из `_TME_LINK_RE`, добавлен invite-парсинг (`t.me/+hash`, `joinchat/hash`) → kind `'invite'`.
- **835#8** `refresh-types`: `resolve_channel(signal_gone=True)` → `{"gone": True}` на definitive not-found; deactivate vs skip (мёртвая `if info is False` убрана) в CLI и worker.

## Тесты (TDD)
parsers (invite), chat_filter (bare-id), dispatcher refresh-types. 240 passed, ruff чисто.

Closes #857 · Part of #835

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #868, opened 2026-06-15
fix(lifecycle): worker-gated recovery, task cleanup, atomic migration (836#2,#10,#11,#7, 838#8)

### LINKED ISSUE #867: PR-14: lifecycle / конкурентность / startup (836#2,#10,#11,#7, 838#8)
Часть аудита #836 и #838.

## Закрываемые баги
- [ ] **836#2** (MED) Рестарт web-контейнера в split-деплое сбрасывал in-flight состояние живого воркера (`recover_running`/`reset_running_on_startup` безусловны). Вся startup-recovery теперь под `if runtime_mode == "worker"`.
- [ ] **836#10** (LOW) Stats-fan-out: `asyncio.gather` без отмены оставлял осиротевшие воркеры при исключении. Теперь `asyncio.TaskGroup` (исключение отменяет sibling'ов, всплывает ExceptionGroup).
- [ ] **836#11** (LOW) `ClientPool.disconnect_all` не отменял `_warming_task`/`_dialog_refresh_tasks`. Теперь отменяет + `gather(return_exceptions=True)` + очистка.
- [ ] **838#8** (LOW) `release_client` снимал `stack[-1]` (LIFO) независимо от типа lease. Теперь приоритетно снимает `disconnect_on_release=True`-lease (эфемерный native-клиент рвётся сразу).
- [ ] **836#7** (LOW) Легаси-перестройка `collection_tasks` не атомарна под autocommit (DROP/RENAME). Окно DROP→RENAME обёрнуто в `BEGIN IMMEDIATE`/`COMMIT`/`ROLLBACK`.

## TDD
`test_client_pool_lifecycle.py` (release prefers disconnect-on-release / LIFO fallback / disconnect_all cancels bg tasks); `test_bootstrap.py` (web-mode skips recovery / worker-mode runs it). 19 passed; regression 175 passed (1 пропуск — отсутствует optional `jieba` в окружении).

Part of #836, #838.

axisrow:
## Что
- **836#2** startup-recovery теперь только под `runtime_mode == "worker"` (web-рестарт не корраптит живой воркер).
- **836#10** stats fan-out: `gather` → `asyncio.TaskGroup` (отмена sibling'ов, нет осиротевших задач).
- **836#11** `disconnect_all` отменяет `_warming_task`/`_dialog_refresh_tasks`.
- **838#8** `release_client` приоритетно снимает `disconnect_on_release`-lease.
- **836#7** окно DROP→RENAME миграции `collection_tasks` обёрнуто в `BEGIN IMMEDIATE`/`COMMIT`.

## Тесты (TDD)
`test_client_pool_lifecycle.py`, `test_bootstrap.py` (web/worker recovery). 19 passed; regression 175 passed (1 пропуск — отсутствует optional `jieba` в окружении, не связано). ruff чисто.

Closes #867 · Part of #836, #838

### DIFF
--- a/src/database/migrations.py
+++ b/src/database/migrations.py
@@ -316,8 +316,17 @@ async def _rebuild_collection_tasks_if_channel_id_notnull(db: aiosqlite.Connecti
         FROM collection_tasks
         """
     )
-    await db.execute("DROP TABLE collection_tasks")
-    await db.execute("ALTER TABLE collection_tasks_tmp RENAME TO collection_tasks")
+    # The connection runs in autocommit (isolation_level=None), so without an
+    # explicit transaction a crash between DROP and RENAME would lose the table
+    # permanently. Make the swap atomic (audit #836/7).
+    await db.execute("BEGIN IMMEDIATE")
+    try:
+        await db.execute("DROP TABLE collection_tasks")
+        await db.execute("ALTER TABLE collection_tasks_tmp RENAME TO collection_tasks")
+    except Exception:
+        await db.execute("ROLLBACK")
+        raise
+    await db.commit()
     logger.info("Migrated collection_tasks: removed NOT NULL from channel_id")
 
 
--- a/src/database/migrations.py
+++ b/src/database/migrations.py
@@ -258,69 +258,77 @@ async def _rebuild_collection_tasks_if_channel_id_notnull(db: aiosqlite.Connecti
         else "CASE WHEN channel_id = 0 THEN 'stats_all' ELSE 'channel_collect' END"
     )
 
+    # Clean up any orphan tmp left by a previous crashed run (this is throwaway
+    # scratch data, not the live table — safe to drop outside the transaction).
     await db.execute("DROP TABLE IF EXISTS collection_tasks_tmp")
-    await db.execute("""
-        CREATE TABLE collection_tasks_tmp (
-            id INTEGER PRIMARY KEY,
-            channel_id INTEGER,
-            channel_title TEXT,
-            channel_username TEXT,
-            task_type TEXT NOT NULL DEFAULT 'channel_collect',
-            status TEXT DEFAULT 'pending',
-            messages_collected INTEGER DEFAULT 0,
-            error TEXT,
-            note TEXT,
-            run_after TEXT,
-            payload TEXT,
-            parent_task_id INTEGER,
-            created_at TEXT DEFAULT (datetime('now')),
-            started_at TEXT,
-            completed_at TEXT
-        )
-        """)
-    await db.execute(
-        f"""
-        INSERT INTO collection_tasks_tmp (
-            id,
-            channel_id,
-            channel_title,
-            channel_username,
-            task_type,
-            status,
-            messages_collected,
-            error,
-            note,
-            run_after,
-            payload,
-            parent_task_id,
-            created_at,
-            started_at,
-            completed_at
-        )
-        SELECT
-            {expr("id")},
-            {channel_id_expr},
-            {expr("channel_title")},
-            {expr("channel_username")},
-            {task_type_expr},
-            {expr("status", "'pending'")},
-            {expr("messages_collected", "0")},
-            {expr("error")},
-            {expr("note")},
-            {expr("run_after")},
-            {expr("payload")},
-            {expr("parent_task_id")},
-            {expr("created_at", "datetime('now')")},
-            {expr("started_at")},
-            {expr("completed_at")}
-        FROM collection_tasks
-        """
-    )
-    # The connection runs in autocommit (isolation_level=None), so without an
-    # explicit transaction a crash between DROP and RENAME would lose the table
-    # permanently. Make the swap atomic (audit #836/7).
+
+    # The connection runs in autocommit (isolation_level=None). The ENTIRE rebuild —
+    # snapshot copy, drop, rename — must run under one BEGIN IMMEDIATE write lock,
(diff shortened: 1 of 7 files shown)


axisrow:
## 🔍 Local review — cycle 1 (Claude subagent + Codex companion)

Reviewed locally, no GitHub bots pinged. Completeness vs #867: **fully implements** (worker-gated recovery, TaskGroup fan-out, disconnect_all cancellation, release lease priority, atomic DROP→RENAME).

|---|---|---|---|
| **FIX** (critical) | claude (Codex didn't flag) | TaskGroup wraps a worker's `DatabaseBusyError` in an ExceptionGroup that bypasses the dispatcher's `except DatabaseBusyError` requeue → STATS_ALL marked permanently FAILED + traceback flood on a transient lock | `task_handlers/stats.py:248` |
| **FIX** (critical) | codex (Claude didn't flag) | `release_client(phone)` scanned the lease stack for any disconnect_on_release lease and popped it → a direct caller releasing could backend-disconnect a native lease still in use by another caller (session closed mid-operation) | `client_pool.py:810` |
| SKIP (minor) | codex (crit) | collection_tasks rebuild copies outside the write txn (lost-write window) | `migrations.py:261` |

**Both criticals FIXed in `1e0dfdc`:**
- TaskGroup wrapped in `except* DatabaseBusyError` → re-raises a plain DatabaseBusyError so the requeue path fires (keeps sibling-cancellation). ExceptionGroup-unwrap contract test added.
- `release_client` reverted to strict LIFO (`stack.pop()`): a native lease acquired last is on top and still torn down promptly (#838/8 leak target), but a native lease buried under a live direct lease is never wrongly released. Regression test for [native, direct] added; updated the test that encoded the wrong-release.

**Migration lost-write = SKIP:** the copy-outside-txn window is mechanically real but **pre-existing** (not introduced by this PR — `git show main` has the same structure; the PR only added the BEGIN IMMEDIATE around DROP→RENAME) and the residual risk is negligible (one-time first-upgrade migration). Out of scope for this PR.

**This PR is the clearest case for two reviewers:** the two confirmed criticals were found by *different* reviewers (Claude → ExceptionGroup, Codex → lease) with essentially zero overlap.

(Local mode: review-only — fixes pushed, no auto-merge.)

axisrow:
## 🔍 Local review (cycle 1)
Reviewed locally (Claude subagent + Codex companion), no bots pinged. All 5 deliverables of #867 verified present (838#8 is strict-LIFO by deliberate prior-round design, not the issue's literal "prefer" wording — correct and locked by tests).

|---|---|---|---|
| FIX | codex | The `collection_tasks` rebuild snapshots the table (`INSERT INTO tmp SELECT FROM collection_tasks`) **before** `BEGIN IMMEDIATE`. Under autocommit in a split web/worker deploy, the web process can INSERT a new task into `collection_tasks` in the window between the snapshot and the transaction; the migration then DROPs the live table and renames the stale tmp, **permanently losing the interleaved task**. Ironic for a lifecycle-hardening PR. | src/database/migrations.py:261-325 |
| SKIP | claude | stats.py TaskGroup: if two workers fail concurrently with DIFFERENT types (one DatabaseBusyError + one other), the re-raised plain DBE merges back into an ExceptionGroup → bypasses the dispatcher's DBE-requeue → FAILED on a partly-transient lock. Edge-case; not a regression (pre-PR gather was already non-deterministic for mixed failures). | src/services/task_handlers/stats.py:253-260 |

### Why the Codex finding is a FIX (Claude rated the area safe)
Claude verified the *crash*-between-INSERT-and-BEGIN case (orphan tmp, cleaned next run — true). But it did **not** consider a **concurrent writer** in the same window. Codex is right: the snapshot is taken outside the write lock, so a parallel web-process INSERT after the SELECT but before `BEGIN IMMEDIATE` is lost when the swap happens. The fix is to take `BEGIN IMMEDIATE` BEFORE creating/populating tmp and hold it through the whole copy→drop→rename→commit, so no writer can interleave.

**Triage of cycle 1/3: 1 FIX, 1 SKIP.** Working in an isolated worktree (a parallel session shares the main checkout).

axisrow:
## ✅ Local review complete (2 cycles)
- **Cycle 1:** 1 FIX — the `collection_tasks` rebuild snapshotted outside the write lock (concurrent-writer data-loss window in split deploy). Fixed by moving `BEGIN IMMEDIATE` before the whole copy→drop→rename (`995a9b6`); +multi-row retention & idempotent-rerun tests.
- **Cycle 2:** 0 FIX — Codex `approve`, Claude subagent reproduced the two-connection race and confirmed the window is closed; BEGIN/DDL-in-txn/ROLLBACK all valid on the autocommit connection. No regression in the other 4 deliverables.
- **Cleanup pass:** the one remaining minor (stats.py TaskGroup `except*` mixed-exception unwrap → `ExceptionGroup` can bypass the DBE-requeue) is **left as-is by choice**: it needs a *simultaneous* mixed-type failure in one TaskGroup window (rare — pure lock contention is all-DBE, which works), it's not a regression (pre-PR `gather` was already non-deterministic for mixed failures), and a correct `except*` rewrite would add real complexity for a rare path. Documented here rather than forced.
- All 5 deliverables of #867 verified. ruff clean; migrations + lifecycle + bootstrap + stats suites green.

Note for merge: this branch and the already-merged #858 both touch `src/telegram/client_pool.py` (different methods) — a trivial merge may need a rebase, no logical conflict.

Local mode is review-only on merge — merge is yours to trigger once CI is green.

PR #876, opened 2026-06-17
fix(web): hide unavailable search modes + lazyload heavy analytics pages (#833, #756)

axisrow:
## Summary

Two milestone v0.2.4 UX fixes that surfaced on the production database.

### #833 — Search radio buttons no longer dead
Unavailable search modes (semantic/hybrid without a semantic backend; telegram modes without a connected account) were rendered `disabled`. On prod `sqlite-vec` never loads, so they stayed permanently dead and confusing.

- Template now **hides** unavailable modes via a single `available_modes` set (one source of truth shared by the template gate and the handler), instead of `disabled`/`text-muted`.
- The handler normalises a mode reachable only via a hand-crafted URL/POST back to `local`, so the page renders a normal local search rather than a scary error banner.

### #756 — Lazyload heavy analytics pages
`/analytics` and `/calendar` ran 3–4 heavy aggregations **before** rendering, blocking TTFB (~107s on the 45 GB prod DB — effectively a blank screen).

- Pages now paint a **skeleton instantly** and load each heavy section as an HTMX fragment (`hx-get` + `hx-trigger="load"`), following the existing `_debug_logs.html` fragment pattern.
- New fragment routes (`/analytics/fragments/*`, `/calendar/fragments/*`) + partial templates under `analytics/` and `calendar/`.
- Auth is inherited from the global middleware — new routes need no per-route wiring.
- Shared `loading()` macro added to `_macros_ui.html`.

#### Measured on the real 45 GB DB
| | Before | After |
|---|---|---|
| `/analytics` TTFB | ~107 s (blank) | **0.017 s** (skeleton) |
| sections | blocked the whole page | stream in independently (4 s / 39 s / 64 s) |
| `/calendar` TTFB | sum of 4 queries | **0.006 s** |

The slow fragments (engagement 39 s, hourly 64 s) are the full-scan/COUNT queries that #760 (DB-perf) will address next — lazyload removes the UI block now.

## Test plan
- Updated 4 tests whose assertions moved into fragments (now assert against the fragment endpoints).
- Added tests: search modes hidden + local fallback; analytics/calendar skeletons expose HTMX markers; fragment endpoints return bare partials (no `<html>`).
- `ruff` clean; full web/route suite green (1166 tests).

Both passes went through `/simplify` review (reuse/simplification/efficiency/altitude).

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -54,6 +54,9 @@ class NotificationsConfig(BaseModel):
 
 class DatabaseConfig(BaseModel):
     path: str = "data/tg_search.db"
+    # Number of read connections in the pool (#760). Reads run concurrently across
+    # these, so a slow SELECT never blocks navigation. Ignored for :memory:.
+    read_pool_size: int = 4
 
 
 class LLMConfig(BaseModel):
--- a/src/database/connection.py
+++ b/src/database/connection.py
@@ -72,29 +72,55 @@ class ProfilingConnection:
             profiler.record_db(time.perf_counter_ns() - t0)
 
 
+async def apply_pragmas(conn: aiosqlite.Connection, *, role: str = "write") -> None:
+    """Apply the connection-tuning PRAGMAs shared by every connection (#760).
+
+    ``role="write"`` runs the full set including WAL setup/hygiene; ``role="read"``
+    skips the write-only checkpoint PRAGMAs (a read connection never checkpoints)
+    but keeps the per-connection cache/mmap/temp tuning so pool readers are as fast
+    as the writer.
+    """
+    await conn.execute("PRAGMA synchronous=NORMAL")
+    await conn.execute("PRAGMA foreign_keys=ON")
+    await conn.execute("PRAGMA cache_size=-64000")  # 64MB
+    await conn.execute("PRAGMA temp_store=MEMORY")
+    await conn.execute("PRAGMA mmap_size=30000000")  # 30MB
+    if role == "write":
+        await conn.execute("PRAGMA journal_mode=WAL")
+        # WAL hygiene (#766): make the autocheckpoint threshold explicit and
+        # trim a WAL grown by a previous run. PASSIVE never blocks other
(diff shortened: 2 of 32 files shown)


axisrow:
## 🔍 Local review (cycle 1)

Reviewed locally (Claude subagent + Codex companion), no bots pinged. Read-pool (#760), lazyload (#756), and search-mode hiding (#833) all verified.

|---|---|---|---|
| SKIP | codex | Split-mode telegram search can normalize to local when the snapshot pool is empty (only via hand-crafted POST; worker publishes `connected_phones`, so a connected account keeps the proxy path) | src/web/search/handlers.py:215 |
| SKIP | codex | `_reject_writes` checks only the first SQL token (writable-CTE/VACUUM/ATTACH bypass) — defensive only; no such write exists on the read path today, real direct-DML writes ARE caught | src/database/pool.py:123 |
| SKIP | claude | Length filter dropped on crafted `?mode=telegram&q=foo:50` (edge via crafted URL) | src/web/search/handlers.py:195 |
| IRRELEVANT | claude | facade `.execute()` reads stay on write conn — intended generic escape hatch, pre-existing | src/database/facade.py |
| SKIP | claude | `_ensure_udf` caches a single flag — correct for the fixed-size pool | src/database/repositories/filters.py:118 |

**No FIX-level findings.** The high-risk read-pool refactor is sound: write-lock contract (#569) intact, no connection leak on query error, BufferedCursor closes the cursor-after-release race, `:memory:` degeneration correct, lazyload fragments protected by global auth. The non-blocking edges above will be tightened in the final cleanup pass.

axisrow:
✅ Final cleanup pass applied in 99e9daa — the deferred non-blocking findings (Codex C1: telegram modes reach the worker proxy in web runtime instead of silently falling back to local; Claude Cl1: length filters resolved against the normalised mode). New regression test added; full parallel-safe suite green (8068 passed). Review cycle complete — merge left to the maintainer.

PR #877, opened 2026-06-17
perf(web): lazyload /dashboard and /channels (#756)

axisrow:
## Summary

Continues the lazyload work (#756, after #876) for the two slowest **non-analytics** pages measured on the 45 GB prod DB. Both now paint a skeleton instantly and load their heavy data via HTMX fragments — same pattern as `/analytics`.

### `/dashboard/`
`get_stats()` runs `COUNT(*)` over millions of messages (+ calendar/task/pipeline queries) — all moved to `GET /dashboard/fragments/overview`.
**Onboarding redirects** (no auth → `/settings`; no accounts → `/settings?msg=no_accounts`) stay synchronous, so a skeleton 200 can't swallow them.

### `/channels/` (0.6–9.8 s cold)
`list_channels_with_counts` (JOIN channels+stats) + `get_latest_and_previous_stats` + 2× `count_channels` → `GET /channels/fragments/list`.

**OOB-button care (non-trivial):** the per-row collect buttons swap OOB into both `collect-btn-{pk}` (desktop) and `collect-btn-m-{pk}` (mobile) copies, so **desktop table AND mobile cards live together in the one fragment** — splitting them would break the OOB swap. The independent `collect-all-btn` stays in the skeleton.

## Test plan
- Redirected table-data assertions to the fragment endpoints (`/dashboard/fragments/overview`, `/channels/fragments/list`).
- Added: skeleton has `hx-get` + `hx-trigger="load"` markers; dashboard onboarding redirect preserved; fragments return bare partials; **channels fragment renders both OOB copies**.
- Full parallel-safe suite green; `ruff` clean.

## Scope
This is **wave 1 (part 1)** of the slow-page lazyload. `/scheduler/` and `/dialogs/` (also slow, but with non-trivial gather-context split / existing-HTMX retargeting) follow in a separate PR. `/settings/` is excluded (only ~0.4 s; `settings.js` re-init risk not worth it).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #878, opened 2026-06-17
perf(web): lazyload /scheduler and /dialogs heavy sections (#756)

axisrow:
## Summary

Wave 1b of the lazy-load rollout (#756). Splits the **/scheduler** and **/dialogs** pages so the initial request returns only lightweight controls and defers heavy data to HTMX fragment requests — mirroring the pattern already used for /dashboard and /channels (#877).

### Scheduler
- `/scheduler` skeleton now renders only the action controls. Health, jobs and tasks each load via `/scheduler/fragments/{health,jobs,tasks}`, each with its own concurrent `asyncio.gather`, so a slow collector-health snapshot no longer blocks the jobs/tasks tables.
- Extracted `scheduler/_health.html`, `_jobs.html`, `_tasks.html` partials.

### Dialogs
- `/dialogs` skeleton renders only the account selector. The dialog list (a network/cache fetch) loads via `/dialogs/fragments/list`, so switching accounts no longer reloads the whole page.
- Extracted `dialogs/_list.html` partial.

## Cleanup (from /simplify pass)
- Shared `scheduler_filter_qs()` macro in `_macros_ui.html` replaces the filter/page/limit query-string block that the split had copy-pasted into `scheduler.html` and the `_jobs`/`_tasks` fragments.
- Restored `filter_qs` on the `clear-pending-collect` form — dropped during the partial extraction (sibling cancel forms kept it).
- Dropped an orphaned `_get_command_state` lookup + `command` context from the dialogs list fragment (no template consumes it).

## Testing
- `pytest tests/routes/test_scheduler_routes.py tests/routes/test_dialogs_routes.py` — 169 passed
- `ruff check` clean on changed files
- Verified the `scheduler_filter_qs` macro renders identical query strings to the old inline block (no stray whitespace)


axisrow:
## 🔍 Local review (cycle 1)

Reviewed locally (Claude subagent + Codex companion), no bots pinged. Both claims verified against the actual code before assigning a verdict.

|---|---|---|---|
| FIX | codex | Active-task polling calls `window.location.reload()` — under the highest-load scheduler state it reloads the whole page and re-fires all three lazy fragments (health/jobs/tasks) every 5s, defeating the lazyload split this PR introduces | `scheduler/_tasks.html:176-211` |
| FIX | codex + claude | Refresh form + `hx-push-url="true"` live outside the `#dialogs-list` swap target: after an HTMX account switch the hidden `phone` is stale (refresh posts the wrong account / stays disabled), and the address bar is pushed to the bare `/dialogs/fragments/list?phone=…` URL — a reload/bookmark then renders the chrome-less partial | `dialogs.html:27-44` |

Verified NOT problems (claims that held up): `scheduler_filter_qs` macro renders byte-identical query strings to the old inline block (no whitespace bug); macro imported in every consuming template; heavy work genuinely deferred (skeleton routes dropped the gathers); fragment routes registered + wrapped through the response helpers; `filter_qs` correctly restored on the clear-pending form; `_get_command_state` removal safe.

Applying both FIXes now.

axisrow:
## 🔍 Local review (cycle 2)

Reviewed locally (Claude subagent + Codex companion), no bots pinged. Both reviewers confirmed the two cycle-1 fixes are correct (HTMX tasks self-refresh wires up and stays capped; the relocated dialogs refresh form posts the right phone). Codex surfaced one remaining issue of the same class.

|---|---|---|---|
| FIX | codex | After an HTMX account switch (which swaps only `#dialogs-list`), the **Photo Loader** link — an account-scoped control outside the fragment — keeps the `?phone=` from the initial render, so it can send the user into the wrong account's Photo Loader | `dialogs.html:14` |

Codex rated this `medium` (a navigation link, not a mutating action), but it's the same root defect as the cycle-1 refresh-form fix, so I closed it deeply per Codex's own recommendation ("move/sync every `selected_phone`-dependent control"): the selector now updates the Photo Loader `href` on change via a small JS handler, keeping the out-of-fragment control in sync. Added a render-test assertion.

Claude subagent verdict: **APPROVE** — both cycle-1 fixes correct, no new bugs, no infinite-loop/double-fire in the capped self-refresh.

Cycle 2/3 triage: 1 FIX (applied), 0 SKIP. Applying and re-reviewing.

axisrow:
## ✅ Local review complete — cycle 3/3 (final)

Both reviewers (Claude subagent + Codex companion) returned **APPROVE** with zero findings this round. All three regressions found across the cycle are fixed and verified:

- **`5a…` cycle 1** — scheduler active-task polling self-refreshes only `#tasks-fragment` via HTMX (no more full-page `window.location.reload()` that re-fired every lazy fragment), with the bounded-retry cap preserved; dialogs refresh form moved into the swapped fragment so its phone always matches the shown account, `hx-push-url` dropped.
- **`24d147e` cycle 2** — Photo Loader link (the one remaining account-scoped control outside the swapped fragment) kept in sync with the selected account via a null-safe, properly-encoded JS handler.

Codex cycle-3: *"Ship — no remaining `selected_phone`-dependent controls outside the swapped fragment; HTMX wiring preserves the intended targets; no critical data-loss, security, or stale-account path."*
Claude cycle-3: **APPROVE** — stale-control sweep confirms Photo Loader was the only such control and it's fixed.

### Cleanup pass (step 6.5)
One non-blocking minor was noted (`left`/`failed` query params on the dialogs lazy `hx-get` are vestigial), but it **pre-dates this PR** — the leave flow went fully async via `CommandRedirect` before this change, so that confirmation message has been dead since then, and a route test (`test_dialogs.py:360`) deliberately covers the fragment still accepting those params. Removing it spans the route/handler/fragment signature and is out of this PR's scope; left for a dedicated follow-up cleanup.

**Tests:** 223 passed across the scheduler/dialogs/photo-loader route + template suites. `ruff check` clean.

Review mode was **local** — this loop does not auto-merge. Merge is yours to trigger whenever you're ready (CI on the branch can be confirmed first).

PR #945, opened 2026-06-22
perf(db): #760 database performance — FTS backfill, drop 21GB index, reactions date, ANALYZE seed

axisrow:
Closes the remaining work on #760 (DB performance on large databases). Investigation showed the reported "search is slow" was actually **analytics** — the trending-emojis fragment took **3m43s** on a 6.8M-row reactions table. Four independent fixes, all gated/idempotent migrations, verified by measurement on a real 48 GB / 25.6M-row local database.

## What changed

**1. FTS5 backfill migration** (`9cb468c`)
`messages_fts` is kept in sync only by INSERT/DELETE triggers, so a database rebuilt from a dump (or one that predates the FTS table) boots with an empty index — search then silently falls through to a slow `LIKE` scan. Adds an idempotent, state-gated rebuild.

**2. Drop the obsolete ~21 GB `idx_messages_text`** (`34f74d6`)
A B-tree on `messages(text)` cannot serve `LIKE '%..%'` and FTS goes through `messages_fts`, not this column. On the 48 GB DB it weighed **~21 GB** (≈half the file) and slowed every insert. A measured A/B showed it saving only **~1.5%** on the rarely-used LIKE fallback (5m03s vs 5m08s). Dropped via idempotent migration. Freed pages need a manual `VACUUM` (left to the operator).

**3. Denormalize `date` onto `message_reactions`** (`56104b7`) — the headline fix
Trending-emojis filtered by period on `messages`, forcing a full scan of all 6.8M reactions + a JOIN. Adding the parent message's `date` to each reaction (filled on insert; one-off backfill for legacy rows) lets the query filter reactions directly via `idx_message_reactions_date_emoji` and drop the JOIN. On a real 6.8M-row prototype: **3m43s → ~4ms**.

**4. Seed planner statistics with a one-off ANALYZE** (`0e65921`)
`PRAGMA optimize` on close (#940) only re-analyzes tables that changed since the last ANALYZE — on a DB where ANALYZE never ran, `sqlite_stat1` stays empty and the planner is blind. Confirmed empirically on a live DB. Adds a settings-gated, `analysis_limit`-bounded, best-effort ANALYZE.

## Scope

Minimal: only the one slow query was touched. The other reaction queries measured fast (0.07s and below) and are left untouched. No caching/snapshot/precompute — the cause is fixed in the data, not hidden behind a cache.

## Migrations on first start (all gated/idempotent, run once)
- FTS backfill — no-op if the index is already populated.
- `message_reactions.date` backfill — `UPDATE` over existing rows (~54s on the 6.8M-row local DB).
- One-off ANALYZE — bounded by `analysis_limit`.

## Tests
116 passed, ruff clean. New coverage in `tests/test_migrations.py` (FTS backfill, index drop, reactions-date backfill, ANALYZE seeding) and `tests/test_database.py` (date stored on both insert paths).


axisrow:
## 🔍 Local review (cycle 1)
Reviewed locally (Claude subagent + Codex companion), no bots pinged.

|---|---|---|---|
| FIX | codex | date-emoji index built before the 6.8M-row backfill UPDATE → index write-amplification on first boot | src/database/migrations.py (index creation order) |
| FIX | claude | get_trending_emojis SQL only covered by a mocked unit test → headline fix not exercised on a real DB | tests/test_trend_service.py |
| SKIP | claude | duplicate unoptimized MessagesRepository.get_trending_emojis (old JOIN path) — out of scope; addressed in cleanup | src/database/repositories/messages.py:1234 |
| SKIP | claude | trending-emojis incomplete until backfill finishes (NULL dates excluded) | intentional synchronous backfill; correct by design |
| SKIP | claude | one-off ANALYZE is a no-op on a fresh empty DB | limitation, not a bug; #760 targets existing large DBs |

**Fixes applied:**
- Index `idx_message_reactions_date_emoji` is now created by `_backfill_message_reactions_date` **after** the backfill UPDATE (removed from SCHEMA_SQL and SCHEMA_REPAIR_INDEXES), so the 6.8M-row UPDATE no longer pays per-row index maintenance. Regression test added.
- Added an end-to-end `TrendService.get_trending_emojis` test against a real DB (`test_database.py`) — exercises the actual `mr.date` SQL, not just the Python mapping.

117 tests green, ruff clean.

axisrow:
## 🔍 Local review (cycle 2)
Reviewed locally (Claude subagent + Codex companion), no bots pinged.

|---|---|---|---|
| FIX | codex | Index creation gated behind the backfill flag → a DB with the flag but no index (version skew) would never get it recreated | src/database/migrations.py:484 |
| OK | claude | Cycle-1 fixes verified correct against code + git history; index guaranteed in all paths; no regressions | — |

**Fix applied:** the `CREATE INDEX IF NOT EXISTS idx_message_reactions_date_emoji` now runs **unconditionally** every boot (not gated by the backfill flag) — only the expensive UPDATE stays flag-gated. It still runs after the UPDATE, so no write amplification. Added a regression test (`test_reactions_date_index_recreated_when_flag_set_but_index_missing`) that drops the index, sets the flag, and asserts `run_migrations` recreates it.

118 tests green, ruff clean.

Note: Claude and Codex disagreed here — Claude proved no *current* path produces "flag without index"; Codex flagged the *structural* fragility. Applied the fix anyway: ungating the index is cheap, idempotent, and makes the code defensively correct.

axisrow:
## ✅ Local review complete (3 cycles)

**Cycle 1** — 2 FIX applied: index write-amplification (Codex), missing end-to-end test (Claude).
**Cycle 2** — 1 FIX applied: index creation ungated from the backfill flag so it can't be permanently missing (Codex).
**Cycle 3** — both reviewers approve, no FIX. Codex: *"Ship"*. Claude: *"можно мержить"*. One minor (guards before flag = +3 metadata queries/boot) — an intentional trade-off, not a defect.

**Final cleanup pass** — removed the `.beads/` tracker DB (34 files: Dolt backups, embeddeddolt, LOCK) that had leaked into the branch, and added `.beads/` to `.gitignore`. The PR diff is now code-only.

118 tests green, ruff clean. Local mode is review-only on merge — **merge is yours to trigger** once CI is green.

axisrow:

axisrow:
## 🔍 Cycle-review triage (cloud) — cycle 1

**Reviewers conflict on the FTS backfill fast-path**, so flagging for the author to decide — not auto-fixing or merging.

- **Claude** → *Ship*, called the fast-path reliable, but its reasoning only covers the crash / fully-empty case.
- **Codex (P1)** → the fast-path rebuilds only when the index is **completely empty**; a **partially populated** legacy index leaves historical messages unindexed.

**Independent verification — Codex's P1 holds (confirmed by reading the code):**
- `messages_fts` is created in `SCHEMA_SQL` via `CREATE VIRTUAL TABLE IF NOT EXISTS` + AFTER INSERT/DELETE triggers.
- A DB that held messages **before** the FTS table was introduced: on the upgrade the table is created **empty**, then triggers populate it from **new inserts only** → `messages_fts_docsize` becomes non-empty while pre-existing rows stay out of the index.
- This backfill migration is new, so on such a real state the fast-path (`SELECT 1 FROM messages_fts_docsize LIMIT 1`) sees a non-empty docsize and **returns without repairing** the historical rows.
- `MessagesRepository.search_messages` uses `MATCH` whenever `_fts_available` is true, with **no `LIKE` fallback** → those old messages become unsearchable. This is precisely the scenario fix #1 targets, so it may not repair the real 48 GB DB.

**Why it's left to you, not auto-fixed:** robust partial-coverage detection needs a `COUNT(messages)` vs `COUNT(docsize)` comparison — the millions-row `COUNT(*)` you deliberately avoided on every boot. That's a real correctness/perf trade-off only the author should make. Options: (a) detect partial coverage once (e.g. a bounded/`MAX(rowid)`-style probe rather than a full COUNT) and rebuild; (b) accept the trade-off and document that legacy partial indexes need a manual one-off `INSERT INTO messages_fts(messages_fts) VALUES('rebuild')`.

The other findings checked out clean (migration crash-safety, reaction-date format, ANALYZE seeding, test coverage). No merge performed by cycle-review.

PR #1128, opened 2026-06-24
fix(publish): persist per-target progress after each delivery (#1116)

### LINKED ISSUE #1116: bug(publish): дубль отправки при сбое DB после частичной доставки (HIGH)
**Severity: HIGH — data-loss / дубль отправки в Telegram.** Найдено Codex bug-hunt.

## Локация
`src/services/publish_service.py:93-100`

## Сценарий
Telegram-отправки в цикле по target'ам; каждая успешная пишется в `metadata["published_targets"]`, НО `set_published_at()` вызывается только в конце при всех успешных. Если DB-запись после ЧАСТИЧНОЙ отправки падает → задача FAILED → ретрай начинает заново, включая уже доставленные → дубль.

Пример: 3 target, отправлено в 2, DB после 2-го упала → FAILED → ретрай шлёт во все 3 → первые 2 получают дубль.

```python
if newly_delivered:
    delivered.update(newly_delivered)
    metadata["published_targets"] = sorted(delivered)
    await self._db.repos.generation_runs.set_metadata(run.id, metadata)
...
if all(r.success for r in results):
    await self._db.repos.generation_runs.set_published_at(run.id)
```

## Фикс (TDD)
Ретрай должен пропускать уже-доставленные target'ы (per-target дедуп через published_targets — проверять ПЕРЕД отправкой в каждый target, не только писать после). Красный тест: частичная доставка + сбой DB → ретрай НЕ шлёт в уже-доставленные. Связано #633 (per-target tracking), #1082.

axisrow:
## Problem

**Closes #1116** — HIGH severity, data-loss class (duplicate Telegram sends). Found by Codex bug-hunt.

`PublishService.publish_run` sends to each pipeline target in a loop and records delivered targets in an in-memory `delivered` set, but persisted `metadata["published_targets"]` with a **single** `set_metadata()` call *after* the loop.

A Telegram send is irreversible and there is no transaction spanning the send and the DB write. So if that one end-of-loop write failed after a **partial** delivery (e.g. 2 of 3 targets sent), the run went `FAILED` with **no** progress recorded. The retry then re-sent to the already-delivered targets → **duplicate message in the channel**.

```
3 targets → delivered to 2 → DB write after loop fails → run FAILED
→ retry re-sends to all 3 → first 2 get a duplicate
```

## Fix

Write `published_targets` **immediately after each successful delivery** instead of batching to the end of the loop. This bounds the worst case to re-sending the single in-flight target on retry, rather than every target delivered in the attempt.

The write is **deliberately not** wrapped in `try/except`: a failed progress write means we can no longer remember what we delivered, so the raised exception must stop the loop rather than keep sending to targets we cannot track.

Telegram has no app-level idempotency key, so a duplicate of exactly one in-flight target on a crash-in-the-gap is the irreducible floor (write-after-send). Writing *before* send would instead risk a lost delivery, which is worse for a data-loss issue — so write-after-send is the correct shape. (Root cause and minimal-fix shape independently confirmed by a bug-analyzer pass.)

The pre-send per-target dedup (`if key in delivered: skip`, the #633 layer) is unchanged.

## Tests

Two red tests added and **mutation-verified** (they fail against the old batched-write code, pass against the fix):

1. `test_publish_service_persists_progress_after_each_delivery` — asserts one `set_metadata` write **per** delivered target, not a single end-of-loop write.
2. `test_publish_service_db_failure_after_partial_delivery_no_duplicate_on_retry` — the #1116 scenario end to end: 3 targets, delivered to 2, DB write fails → run FAILED → **retry does NOT re-send** to the already-delivered target.

All existing publish/dispatcher/e2e tests stay green. Full suite: 3322 + 1078 passed (parallel + serial), 0 failures. `ruff` clean, no new `mypy` errors.

Related: #633 (per-target tracking), #1082.

### DIFF
--- a/src/services/publish_service.py
+++ b/src/services/publish_service.py
@@ -95,7 +95,6 @@ class PublishService:
         delivered: set[str] = set(metadata.get("published_targets") or [])
 
         results: list[PublishResult] = []
-        newly_delivered: list[str] = []
         for target in targets:
             key = _target_key(target)
             if key in delivered:
@@ -104,12 +103,19 @@ class PublishService:
                 continue
             result = await self._publish_to_target(run, target)
             results.append(result)
-            if result.success:
-                newly_delivered.append(key)
-
-        # Persist incremental progress so a later retry resumes only the failed targets.
-        if newly_delivered:
-            delivered.update(newly_delivered)
+            if not result.success:
+                continue
+            # Persist progress immediately after EACH delivery — never batched to a
+            # single end-of-loop write (issue #1116). A send to Telegram is
+            # irreversible and there is no transaction spanning send + DB write, so
+            # if this write fails the run goes FAILED and is retried. Recording the
+            # delivered target right now bounds the worst case to re-sending the one
+            # in-flight target on retry; a batched write would instead lose every
+            # target delivered in this attempt and duplicate them all. The write is
+            # deliberately NOT wrapped in try/except: a failed write means we can no
+            # longer remember what we delivered, so the raised exception must stop
+            # the loop rather than keep sending to targets we cannot track.
+            delivered.add(key)
             metadata["published_targets"] = sorted(delivered)
             await self._db.repos.generation_runs.set_metadata(run.id, metadata)
 
--- a/tests/test_publish_service.py
+++ b/tests/test_publish_service.py
@@ -5,14 +5,16 @@ from src.services.publish_service import PublishService
 
 
 class FakeDB:
-    def __init__(self):
-        self.repos = FakeRepos()
+    def __init__(self, fail_set_metadata_after=None):
+        self.repos = FakeRepos(fail_set_metadata_after=fail_set_metadata_after)
 
 
 class FakeRepos:
-    def __init__(self):
+    def __init__(self, fail_set_metadata_after=None):
         self._pipelines = FakePipelinesRepo()
-        self._generation_runs = FakeGenerationRunsRepo()
+        self._generation_runs = FakeGenerationRunsRepo(
+            fail_set_metadata_after=fail_set_metadata_after
+        )
 
     @property
     def content_pipelines(self):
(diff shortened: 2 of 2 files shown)


axisrow:
## 🔍 Local review (cycle 1)

Reviewed locally (Claude subagent + Codex companion), no bots pinged. Both reviewers converged on **approve** with **zero critical findings**.

|---|---|---|---|
| APPROVE | codex | No critical blocker; progress persistence moved to per-success path, metadata write left unhandled so DB failures stop further sends, retry preserved for already-persisted targets | — |
| APPROVE | claude | Fix is minimal, correct, well-tested; closes #1116 to the irreducible 1-target floor; #633 dedup intact; all 4 production callers handle the raised exception → task FAILED | src/services/publish_service.py |
| SKIP→FIX | claude | Second retry test (`fail_set_metadata_after=1`) only proves dedup of the **first** delivered target; the headline "2 of 3 delivered" scenario isn't asserted | tests/test_publish_service.py:759 |

**Verification highlights (Claude subagent, all 5 checks passed):**
- Incremental write shrinks the duplicate window from **N targets → 1** (irreducible: no transaction spans the Telegram send + SQLite write; write-after-send is the correct direction).
- `set_metadata` overwrites the full `metadata` JSON, but `publish_run` writes the complete dict each time (only mutating `published_targets`), so other keys are preserved. No concurrent writer; connection-wide write-lock serializes.
- Exception propagation confirmed across all production callers (`task_handlers/content.py`, `dispatcher/moderation_mixin.py`, `agent/tools/pipeline_runs.py`) → run marked FAILED, never published.
- Red tests are mutation-verified and the dict-aliasing trap is explicitly handled (`set_metadata` stores `dict(metadata)`).

Applying the one minor finding (extend test coverage to the 2-of-3 headline case) in a follow-up cleanup commit before handing back. **Not auto-merging from the review skill** — this is a data-loss fix; merge decision is recorded for the orchestrator.

PR #1270, opened 2026-07-09
fix(pool): stop client-pool eviction/queue-drain on an unavailable phone

axisrow:
Two client-pool defects surfaced by the #1234 code review, both about an *unavailable single phone* being mistaken for a *pool-wide failure*.

## #1242 — force_native failure evicts a healthy pooled client (session leak)

`pool_lifecycle._acquire_from_lease` sets `direct_session = None` unconditionally when `force_native=True` — it uses an ephemeral native client and never touches `self.clients`. But the `except` branch popped `self.clients` whenever `direct_session is None`, which is **always** true under `force_native`. So any error building the ephemeral lease evicted a live pooled client **without disconnecting it** — the pooled session vanishes from the map and its connection leaks.

**Fix:** guard the pop with `not force_native`, so a `force_native` acquisition failure leaves the pooled session untouched. The non-force path (genuinely absent / reconnect-broken direct session, or a fresh non-force acquire that raises) still evicts as before.

## #1245 — one busy preferred phone drains the entire collection queue

`collection._acquire_collection_client`, for a private channel with a known preferred phone, called `get_client_by_phone(preferred)`. If that returned `None` (preferred phone flood-waited / in-use / gone), it fell through to `_raise_collection_unavailability` → `NoActiveCollectionClientsError`. That error cascades into `collection_queue.py`, which calls `_drain_memory_queue()` and wipes **every** other channel's pending task — because a *single* preferred phone was busy.

**Fix:** on `None` from the preferred phone, fall back to `get_available_client()` (skip/rotate to any free account) before raising. A genuine "no clients at all" outage still surfaces `NoActiveCollectionClientsError` (and its requeue-and-defer path); a single busy preferred phone no longer does.

## Tests

Each fix has a regression guard that was **mutation-proven red** (revert the one-line fix → the guard fails), plus a companion guarding the fix from over-reaching:

- `test_acquire_from_lease_force_native_failure_keeps_pooled_client` — force_native failure must not evict the pooled client.
- `test_acquire_from_lease_non_force_fresh_acquire_failure_still_pops` — the non-force path must **still** evict on failure.
- `test_acquire_falls_back_when_preferred_phone_unavailable` — unavailable preferred phone rotates to another account.
- `test_acquire_raises_when_preferred_unavailable_and_no_fallback` — a real no-clients outage still raises.

[...]

PR #1275, opened 2026-07-09
chore(mypy): type telegram & shared utils (src/ 42 -> 30)

axisrow:
Part of #1133 (mypy debt reduction axis).

## mypy delta

`python -m mypy src/`: **42 errors -> 30 errors** (−12). No new errors in untouched files (verified against per-file baseline).

|---|---|---|
| src/settings_utils.py | 2 | 0 |
| src/telegram/resolve_guard.py | 4 | 0 |
| src/telegram/identity.py | 1 | 0 |
| src/telegram/flood_wait.py | 1 | 0 |
| src/telegram/collector.py + collector_mixins/cancellation.py | 2 | 0 |
| src/search/telegram_search.py | 1 | 0 |
| src/utils/datetime.py | 1 | 0 |

- **settings_utils / identity**: targeted `# type: ignore[call-overload|arg-type]` on `int()`/`float()` coerce-and-catch sites — `TypeError` is the handled path in the existing `except`; same pattern as `runtime_diagnostics.py:70`.
- **flood_wait**: `coerce_flood_wait_seconds` param narrowed `object` -> `int | float | str | None`; both callers pass `getattr(exc, "seconds", 0)`.
- **resolve_guard**: TYPE_CHECKING `_SettingsStore` Protocol + `cast` inside `restore_resolve_username_backoff`; the `db: object` signature is kept intact because the pool_lifecycle stub mirrors it verbatim (changing it there is out of scope — #1270 zone).
- **cancellation mixin**: explicit class-level `_last_unavailability_log` declaration matching the runtime `Collector.__init__` annotation; explicit `is None` narrowing before `int(retry_after_sec)` (the helper already returned False for None — identical behavior).
- **telegram_search**: `assert self._pool is not None` before `run_live_username_resolve` — `_acquire_search_client` yields an error result when the pool is missing, so a live client implies a live pool.
- **utils/datetime**: `@overload` set on `normalize_utc` so `parse_required_utc_datetime` returns non-optional `datetime`.

## Verification

- `python -m mypy src/` — 30 errors (was 42), untouched files unchanged
- `pytest tests/test_settings_utils.py tests/test_flood_wait.py tests/test_utils_datetime.py tests/test_client_pool_resolve_guard.py tests/test_collector.py tests/test_collector_internals.py tests/test_search.py` — **203 passed**

### DIFF
--- a/src/search/telegram_search.py
+++ b/src/search/telegram_search.py
@@ -552,6 +552,9 @@ class TelegramSearch:
                         )
                         username = ch_record.username if ch_record else None
                         if username:
+                            # _acquire_search_client yields an error result when the
+                            # pool is missing, so a live client implies a live pool.
+                            assert self._pool is not None
                             try:
                                 entity = await self._pool.run_live_username_resolve(
                                     lambda: session.resolve_entity(username),
--- a/src/settings_utils.py
+++ b/src/settings_utils.py
@@ -14,7 +14,8 @@ def parse_int_setting(
     if raw_value in (None, ""):
         return default
     try:
-        return int(raw_value)
+        # int() raising TypeError on unsupported objects is the handled path below.
+        return int(raw_value)  # type: ignore[call-overload]
     except (TypeError, ValueError):
         logger.warning("Invalid %s in settings DB (%r), using %d", setting_name, raw_value, default)
         return default
@@ -30,7 +31,8 @@ def parse_float_setting(
     if raw_value in (None, ""):
         return default
     try:
-        value = float(raw_value)
+        # float() raising TypeError on unsupported objects is the handled path below.
+        value = float(raw_value)  # type: ignore[arg-type]
         # float("nan")/float("inf") parse without raising; reject them so callers
         # that int()-coerce or clamp the result don't crash or propagate NaN.
         if not math.isfinite(value):
--- a/src/telegram/collector_mixins/cancellation.py
+++ b/src/telegram/collector_mixins/cancellation.py
@@ -32,6 +32,11 @@ logger = logging.getLogger("src.telegram.collector")
 
 
 class CancellationMixin:
+    # Declared here so the runtime Collector's __init__ annotation (which widens the
+    # second element to ``str | int | None`` and allows ``None``) matches the type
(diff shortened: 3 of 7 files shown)


axisrow:
## 🔍 Local review (cycle 1) — PR #1275

Reviewed locally (`/review` + Codex companion `sol`/`high`), no bots pinged.
PR head: `910b27bf` (rebased onto `main` `53001148`).

**Verdicts: 0 FIX, 3 SKIP (all non-blocking), 0 UNVERIFIED. Codex: APPROVE (0 findings).**

|---|---|---|---|
| SKIP | claude | `assert self._pool is not None` relies on `_acquire_search_client` early-return contract; would surface as AssertionError (or AttributeError under `python -O`) only in an unreachable-today path | src/search/telegram_search.py:555 |
| SKIP | claude | Added `retry_after_sec is None or …` is pure type-narrowing — `is_transient_flood_wait_seconds(None)` already returns False, so behavior is identical | src/telegram/collector_mixins/cancellation.py:176 |
| SKIP | claude | `coerce_flood_wait_seconds` param narrowed `object`→`int|float|str|None`; both callers (`flood_wait.py:103`, `collection.py:1171`) fit, static-only | src/telegram/flood_wait.py:37 |
| APPROVE | codex | "Casts, overloads, TYPE_CHECKING declarations, and type ignores do not alter runtime behavior. The new assertion is protected by the client-acquisition invariant." 0 findings. | — |

**Triage verification (per-finding, against the code):**
- All three SKIPs confirmed as behavior-preserving — the PR is typing-only (`cast`/`assert`/`@overload`/`TYPE_CHECKING` Protocol/`type: ignore` with explanatory comments). No runtime path changed.
- Codex's note that "non-finite float rejection is a real runtime change" refers to the **pre-existing** `math.isfinite` block in `settings_utils.parse_float_setting`, not this PR's diff (which only adds `# type: ignore` on the `float(raw_value)` call) — out of scope here.

**Local verification:** `ruff` clean; `mypy src/` 42→30 (−12); targeted `pytest` 203 passed. This is the final cycle (no FIX). Local mode does not merge — merge is yours to trigger.

PR #1277, opened 2026-07-09
chore(mypy): type database layer (src/ 42 -> 29)

axisrow:
Part of #1133 (mypy debt reduction axis). Independent sibling of #1275 (telegram utils) — disjoint files.

## mypy delta

`python -m mypy src/` on this branch: **42 errors -> 29 errors** (−13). No new errors in untouched files (verified against per-file baseline).

|---|---|---|
| src/database/facade.py | 3 | 0 |
| src/database/migrations.py | 2 | 0 |
| src/database/pool.py | 2 | 0 |
| src/database/repositories/collection_tasks.py | 2 | 0 |
| src/database/repositories/messages.py | 2 | 0 |
| src/database/connection.py | 1 | 0 |
| src/database/repositories/filters.py | 1 | 0 |

- **migrations / connection / pool / repositories.filters**: `cast()` around `fetchall()` / `execute_fetchall()` results — aiosqlite stubs declare `Iterable[Row]` while the runtime value is a list; `cast` is a runtime no-op (no copy).
- **facade.transaction()**: bind the asserted connection to a local `db` so the narrowed non-optional type is visible inside the busy-retry lambda (mypy does not propagate attribute narrowing into closures).
- **facade.record_rename_event**: separate `write_cur` variable — the read path returns `BufferedCursor`, the write path `aiosqlite.Cursor` (which owns `lastrowid`).
- **repositories.collection_tasks**: add `ExportTaskPayload` to `_deserialize_payload` return union and `_serialize_payload` param union — both code paths already handle it at runtime (the `task_kind == EXPORT` parse branch and the `isinstance` tuple); this only aligns annotations with existing code. The interop payloads (#960 `DmReplyTaskPayload` etc.) stay excluded per the design note in models.py.
- **repositories.messages**: `is None or == ""` narrowing before `int()` (semantically identical to the tuple-membership check); `assert sq.id is not None` inside the chunk loop — `valid` pre-filters None ids two lines above.

## Verification

- `python -m mypy src/` — 29 errors (was 42), untouched files unchanged
- `pytest tests/repositories/ tests/test_database.py tests/test_database_pool.py tests/test_database_write_lock.py tests/test_migrations.py tests/test_migrations_schema_paths.py` — **496 passed**

### DIFF
--- a/src/database/connection.py
+++ b/src/database/connection.py
@@ -5,6 +5,7 @@ import os
 import time
 from dataclasses import dataclass
 from pathlib import Path
+from typing import cast
 
 import aiosqlite
 
@@ -177,4 +178,5 @@ class DBConnection:
 
     async def execute_fetchall(self, sql: str, params: tuple = ()) -> list:
         assert self.db is not None
-        return await self.db.execute_fetchall(sql, params)
+        # aiosqlite stubs declare Iterable[Row]; at runtime this is a list.
+        return cast(list, await self.db.execute_fetchall(sql, params))
--- a/src/database/facade.py
+++ b/src/database/facade.py
@@ -273,20 +273,22 @@ class Database:
         Not reentrant — nesting would deadlock asyncio.Lock.
         """
         assert self._db is not None
+        # Local binding keeps the narrowed (non-optional) type visible inside the lambda.
+        db = self._db
         async with self._write_lock:
             await self._with_busy_retry(
                 "transaction begin",
-                lambda: begin_immediate(self._db),
+                lambda: begin_immediate(db),
             )
             committed = False
             try:
-                yield self._db
-                await self._db.execute("COMMIT")
+                yield db
+                await db.execute("COMMIT")
                 committed = True
             finally:
                 if not committed:
                     try:
-                        await self._db.execute("ROLLBACK")
+                        await db.execute("ROLLBACK")
                     except aiosqlite.OperationalError:
                         pass
 
@@ -578,7 +580,7 @@ class Database:
         if existing:
             return existing["id"]
         try:
-            cur = await self.execute_write(
+            write_cur = await self.execute_write(
                 """
                 INSERT INTO channel_rename_events
                     (channel_id, old_title, new_title, old_username, new_username)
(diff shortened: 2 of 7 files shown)


axisrow:
## 🔍 Local review (cycle 1) — PR #1277

Reviewed locally (`/review` + Codex companion `sol`/`medium`), no bots pinged.
PR head: `a2fb8133` (rebased onto `main` `53001148`).

**Verdicts: 0 FIX, 5 SKIP (all non-blocking), 0 UNVERIFIED. Codex: APPROVE (0 findings).**

|---|---|---|---|
| SKIP | claude | `transaction()` local `db = self._db` binding — same reference, write-lock semantics (`self._write_lock`, BEGIN/COMMIT/ROLLBACK) untouched; solves mypy closure-narrowing | src/database/facade.py:275 |
| SKIP | claude | `cast("list[...]", fetchall())` — aiosqlite stubs say `Iterable[Row]`, runtime returns a list; cast is a no-op (no `list()` copy, no materialization) | connection.py:182, migrations.py:247, pool.py:172,177 |
| SKIP | claude | `ExportTaskPayload` added to `_deserialize_payload` return + `_serialize_payload` param unions — code already handled EXPORT (deserialize branch L97, serialize isinstance tuple L128); annotations only lagged. Interop payloads (`DmReplyTaskPayload` etc.) intentionally excluded per models.py design note. | collection_tasks.py:69,109 |
| SKIP | claude | `raw_value in (None, "")` → `raw_value is None or raw_value == ""` — identical for `str | None`; pure narrowing before `int()` | messages.py:148 |
| SKIP | claude | `assert sq.id is not None` after `valid = [sq for sq in queries if sq.id is not None]` filter — None unreachable; no `await` between filter and assert, so id can't change | messages.py:1220 |
| APPROVE | codex | "transaction binding preserves write-lock semantics; casts are runtime no-ops over list-returning fetchall; ExportTaskPayload changes only annotations; sq.id cannot change between filtering and assertion because no await intervenes." 0 findings. | — |

**Triage verification (per-finding, against the code):** all 5 confirmed behavior-preserving. PR is typing-only (cast/local-binding/rename/union-alignment/assert with comments). No runtime path changed.

**Local verification:** `ruff` clean; `mypy src/` 42→29 (−13); targeted `pytest` 500 passed (repositories, database, pool, write_lock, migrations). This is the final cycle (no FIX). Local mode does not merge — merge is yours to trigger.

PR #1278, opened 2026-07-09
fix: daily cost cap exceeded by concurrent calls — atomic check-and-reserve (#1250)

### LINKED ISSUE #1250: bug: дневной cost cap превышается конкурентными вызовами (check-then-record без резерва)
Из код-ревью #1234 (находка #13).

## Суть
Дневной cost cap — check-then-record без резервирования: `_lock` сериализует только проверку, а расход бронируется `record_cost` после завершения платного вызова (10-60с) — N конкурентных генераций все проходят проверку. Остаток бюджета = 1 изображение; N генераций стартуют конкурентно (pipeline + agent-tool + веб) → каждая проходит `check_cost_cap` по пре-спенд итогу → все N платных вызовов уходят → `record_cost` бронирует всё постфактум → cap превышен на (N-1)×стоимость. «Жёсткая» гарантия #814 не выполняется.

## Что сделать
- Атомарный check-and-reserve под единым `_lock`: резервировать оценочную стоимость ДО платного вызова, корректировать по факту в `record_cost` (release разницы).
- Тест: N конкурентных генераций при остатке на 1 не превышают cap.

## Файлы
- `src/services/production_limits_service.py:344-349` (check), `record_cost`
- потребитель: `src/services/image_generation_service.py:~142`

## Оценка
- **LOC:** ~110–240 (src ~90 / tests ~120). Claim-гонка + concurrency-тесты + дизайн резерва → +20% и +25%.
- **Размер:** L
- **Время (агент):** ~55–90 мин

axisrow:
Closes #1250. Found in the #1234 code review (finding #13); restores the hard daily-cap guarantee from #814.

## Problem

The daily cost cap was **check-then-record**: `CostTracker._lock` serialized only the *check*, while the spend was booked by `record_cost` only *after* the paid provider call finished (10–60s later). With budget left for exactly one image, N concurrent generations (pipeline + agent-tool + web) all passed `check_cost_cap` against the same pre-spend total → all N paid calls went out → `record_cost` booked everything after the fact → the cap was exceeded by (N−1)×cost.

## Fix: atomic check-and-reserve

`CostTracker` now claims the **estimated** cost under the same lock as the cap check, before the paid call runs, and settles it afterwards:

- **`reserve_cost(tokens, is_image)`** — refreshes the authoritative daily total (#814 cross-instance read preserved), checks `daily + reserved + estimated > cap`, and on success adds the estimate to the in-flight reservation ledger — all under one `_lock` acquisition.
- **`record_cost(...)`** — settles: drops the matching reservation (same deterministic estimate for the same arguments) and books the actual spend in its place. Still safe without a prior reservation (release clamped at zero).
- **`release_cost(...)`** — returns the budget when the paid call failed or never ran, so a failed call can't hold the budget until the day rolls over.
- `check_cost_cap` stays a read-only check but now counts outstanding reservations; `get_remaining_budget()` is net of reservations; `get_stats()` exposes `cost.reserved`.

Callers follow the *reserve → try paid call → settle exactly once* pattern:

- `ProductionLimitsService.acquire()` reserves (and **releases on rate-limit timeout**, where the paid call never runs).
- `execute_with_retry()` releases on a failed attempt (including `CancelledError`) before the retry re-acquires — no reservation pile-up, and no new retry added around the paid call itself.
- `ImageGenerationService.generate()` settles in a `finally` block: success → `record_cost`; error / timeout / empty result / cancellation → `release_cost`.

Reservations are in-memory (per process — which is where the race lives: web + embedded worker share one process by default). Cross-instance protection remains at the recorded-spend level exactly as in #814.

## Tests (mutation-verified)

Red proven on the pre-fix code by stashing the src changes:

- `test_concurrent_generations_do_not_exceed_cost_cap` — 5 concurrent `generate()` calls with budget for one image, adapters parked in-flight on an event: **fails `assert 5 == 1` on the old code** (all 5 paid calls went out), passes with the fix (1 paid call, cap intact).
- `test_concurrent_acquire_admits_only_budgeted_calls` — service-level: exactly one of 5 concurrent acquires is admitted.
- `test_failed_generation_releases_reserved_budget[oserror|timeout|returns_none]` — a failed paid call releases the reservation and a follow-up generation still fits the budget.
- `test_acquire_rate_timeout_releases_cost_reservation`, `test_execute_with_retry_releases_reservation_on_failed_attempt`, `test_execute_with_retry_exhausted_leaves_no_reservation`, plus `reserve/release/settle` unit tests on `CostTracker` (no double-count on settle).

## Verification

- `tests/test_production_limits_service.py` + `tests/test_image_generation_service.py` — 110 passed (incl. 12 new)
- `tests/test_regression_services_coverage.py` — 82 passed; dependent image suites (agent tools / routes / CLI) — 98 passed; `pytest -m smoke` — green
- mypy on both changed files: no new errors vs origin/main

https://claude.ai/code/session_019R5KeKC8yEV89hWy6ndCWa

### DIFF
--- a/src/services/image_generation_service.py
+++ b/src/services/image_generation_service.py
@@ -118,6 +118,7 @@ class ImageGenerationService:
                     retryable=True,
                 )
                 return None
+        result: str | None = None
         try:
             result = await adapter(text, model_id)
         except TimeoutError as exc:
@@ -125,7 +126,6 @@ class ImageGenerationService:
             self._last_failure = ImageGenerationFailure(
                 kind="timeout", provider=provider_name, model=model, message=str(exc),
             )
-            return None
         except Exception as exc:
             # OSError is an expected I/O failure — log calmly without a traceback;
             # anything else is unexpected and warrants the full stack.
@@ -136,10 +136,18 @@ class ImageGenerationService:
             self._last_failure = ImageGenerationFailure(
                 kind="error", provider=provider_name, model=model, message=str(exc),
             )
+        finally:
+            # Settle the budget reserved by acquire() exactly once (#1250): a paid
+            # generation books its cost; a failed/empty one (including cancellation,
+            # which skips the except clauses above) returns the reservation so the
+            # budget isn't held until the day rolls over.
+            if limits is not None:
+                if result:
+                    await limits.record_cost(is_image=True)
+                else:
+                    await limits.release_cost(is_image=True)
+        if result is None:
             return None
-        # The paid generation succeeded → record its cost against the daily cap.
-        if limits is not None and result:
-            await limits.record_cost(is_image=True)
         s3 = getattr(self, "_s3", None)
         if result and s3 is not None:
             if result.startswith("http"):
--- a/src/services/production_limits_service.py
+++ b/src/services/production_limits_service.py
@@ -163,6 +163,13 @@ class CostTracker:
     def __init__(self, config: CostConfig | None = None, db: "Database | None" = None):
         self._config = config or CostConfig()
         self._daily_cost = 0.0
+        # Estimated cost claimed by in-flight paid calls: added by reserve_cost
+        # BEFORE the provider call and removed by record_cost/release_cost after
+        # it, so concurrent callers can't all pass the cap check against the same
+        # pre-spend total (#1250). In-memory only — reservations guard concurrency
+        # within this process; cross-instance protection stays at the recorded
+        # spend level (#814).
+        self._reserved = 0.0
         self._day_start = time.time()
         self._lock = asyncio.Lock()
         # When a DB is provided, the daily cost is persisted to the settings table and
@@ -313,12 +320,24 @@ class CostTracker:
             return self._config.cost_per_image
         return (tokens / 1000) * self._config.cost_per_1k_tokens
 
+    async def _refresh_daily_cost(self, now: float) -> None:
+        """Bring ``self._daily_cost`` up to date. Must be called under self._lock."""
+        if self._db is not None:
+            # Re-read the authoritative total so a sibling instance's spend is
+            # seen before we approve another call against the shared cap (#814).
+            self._daily_cost = await self._read_persisted_cost(now)
+            self._loaded = True
+            return
+        await self._ensure_loaded()
+        if self._maybe_reset_day(now):
+            await self._persist()
+
     async def check_cost_cap(
         self,
         tokens: int = 0,
         is_image: bool = False,
     ) -> tuple[bool, float]:
-        """Check if request is within cost cap.
+        """Check if request is within cost cap (read-only, reserves nothing).
 
         Args:
             tokens: Number of tokens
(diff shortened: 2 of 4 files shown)


axisrow:
## 🔍 Local cycle-review complete — 3 cycles (dual Claude + Codex)

Review mode: **local** (`/review` + Codex companion). No bots pinged. No auto-merge.

### Cycle 1 (head `ed085af8`)
Codex (sol/high) found **3 cancellation-leak paths** in the original atomic-reserve fix:
1. **[critical]** `acquire()` leaked the reservation when cancelled during the rate-limit wait (called before the caller's try/finally).
2. **[critical]** `_reserved` survived the daily rollover (`_maybe_reset_day` cleared only `_daily_cost`).
3. **[critical]** `record_cost` settlement was not cancellation-safe on the DB path.

→ Fixed in `79f0a8b3`. F1 (multi-instance) and F4 (estimate≠actual) triaged **out of scope** (in-memory by design per #1250; pre-existing estimate behavior).

### Cycle 2 (head `79f0a8b3` → `92a8305a`)
Codex found the cycle-1 settlement fix still leaked: on cancellation it dropped the reservation (and released `_lock`) the moment the outer task was cancelled, while the `asyncio.shield`-detached booking was still in flight — a concurrent `reserve_cost` could then observe the old spend with no reservation and admit beyond the cap.

→ Fixed in `92a8305a`: await the detached booking to completion under the lock before touching the ledger. Mutation-verified (`test_record_cost_holds_lock_and_reservation_until_booking_commits`).

### Cycle 3 (head `92a8305a` → `ea97349b`)
Codex found a **variant of one root cause**: after the first CancelledError, the cycle-2 fix awaited `booking` WITHOUT shield, so a **second** `task.cancel()` propagated into the booking wait and aborted settlement — lock released with paid spend unrecorded and reservation stranded.

→ Root fix in `ea97349b`: **cancellation-immune settlement via a shield-loop** — await the booking behind `asyncio.shield` in a loop that absorbs any number of outer cancellation requests (the booking task is never cancelled, so it always commits), then drop the reservation and re-raise a single CancelledError once the spend is durably recorded.

**Diagnosis (3-cycle cap fired):** all three findings were variants of one invariant — *the `(_daily_cost, _reserved)` ledger must stay consistent across every cancellation point*. Cycles 1–2 patched symptoms point by point; cycle 3's root fix closes the single-, double-, and N-cancel variants at the invariant level.

### Verification
- 219 passed across production_limits + image_gen + regression + agent-tools suites
- Mutation-verified: cycle-1 (F2/F2b) and cycle-2 window tests turn red on the pre-fix code

### Open / hand-back
- **Codex companion timeouts:** xhigh and high both exceeded the 600s turn budget; only `effort=medium` completed. Flagged as a systemic companion issue (large concurrency analysis + high effort).
- **Double-cancel regression test:** a deterministic test for the N-cancel variant could not be constructed — the orphan booking converges the ledger in the background, so the bug lives only in the transient window the lock-held test already guards. The shield-loop closes it constructively. A second reviewer may want to pressure-test the shield-loop reasoning.
- **Out of scope (follow-up issues, not blocking #1250):** multi-instance reservation bypass (in-memory by design), actual-spend settlement vs estimate (pre-existing).

**Final head:** `ea97349b`. Not merging — local mode, awaiting owner decision on the shield-loop approach and whether to ship vs add a more robust double-cancel harness.

PR #1280, opened 2026-07-10
chore(mypy): type services/runtime/misc (src/ 42 -> 27)

axisrow:
Part of #1133 (mypy debt reduction axis). Independent sibling of #1275 (telegram utils) and #1277 (database) — disjoint files.

## mypy delta

`python -m mypy src/` on this branch: **42 errors -> 27 errors** (−15). No new errors in untouched files (verified against per-file baseline). The remaining 27 are the PR1 (#1275) and PR2 (#1277) files, which this branch does not touch.

|---|---|---|
| src/agent/tools/_registry.py | 3 | 0 |
| src/config.py | 2 | 0 |
| src/models.py | 2 | 0 |
| src/search/numpy_semantic.py | 2 | 0 |
| src/scheduler/service.py | 2 | 0 |
| src/runtime/worker.py | 3 | 0 |
| src/services/notification_matcher.py | 1 | 0 |

- **pyproject mypy overrides**: add `regex` / `regex.*` to the `ignore_missing_imports` module list (third-party, ships no stubs) — clears the `import-untyped` in `notification_matcher`. Same pattern as the existing `sklearn.*` / `yaml` entries.
- **numpy_semantic**: `TYPE_CHECKING`-import `numpy as np` and `NearestNeighbors`, annotate the lazy `_matrix: np.ndarray | None` / `_index: NearestNeighbors | None` fields (the imports stay runtime-lazy).
- **scheduler/service**: rename the loop-bound `all_active` to `active_pipelines` in the pipeline branch — the previous reuse of the same name (typed `list[SearchQuery]` from the search-query branch above) caused a cross-branch assignment error.
- **runtime/worker**: annotate `bot_payload: dict[str, Any]` — it was inferred as bool-only from the initial `{"configured": False}` then widened with str/int/None values.
- **models.PipelineGraph.from_json**: narrow the param to `str | dict[str, Any]` and `assert isinstance(data, dict)` after `json.loads` so the `.get("nodes")` / `.get("edges")` calls type-check (json.loads returns `Any`).
- **config**: targeted `# type: ignore[call-arg]` on `DatabaseConfig()` — the pydantic mypy plugin mis-reports the `Field(gt=0)` / `Field(ge=0)` defaults as required (the defaults are valid and present).
- **agent/tools/_registry**: the three pool helper functions (`connected_phones_from_pool`, `_pool_reports_connections`, etc.) are intentionally duck-typed over `object | None` with `getattr` + try/except. Replaced the few direct `.clients` / `.connected_phones` attribute accesses (which mypy rejects on `object`) with `getattr(...)` and one `cast(Any, ...)` where the attribute is guarded by `callable(type(...).attr)`.

## Verification

- `python -m mypy src/` — 27 errors (was 42), untouched files unchanged; all 7 PR3 files at 0
- `pytest tests/test_notification_matcher.py tests/test_agent_tools_registry.py tests/test_pipeline_graph.py tests/test_jobs_read_model.py` — **143 passed**
- `pytest tests/test_cli_dialogs_notifications_scheduler.py -k "scheduler or potential_job or job"` — **46 passed**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -174,6 +174,8 @@ module = [
     "langdetect",
     "openpyxl",
     "openpyxl.*",
+    "regex",
+    "regex.*",
     "sklearn.*",
     "telethon",
     "telethon.*",
--- a/src/agent/tools/_registry.py
+++ b/src/agent/tools/_registry.py
@@ -268,7 +268,9 @@ def connected_phones_from_pool(client_pool: object | None) -> set[str]:
     if isinstance(instance_attrs, dict) and "connected_phones" in instance_attrs:
         connected_phones = instance_attrs["connected_phones"]
     elif callable(getattr(type(client_pool), "connected_phones", None)):
-        connected_phones = client_pool.connected_phones
+        # type(client_pool).connected_phones is callable (a property/method); access it,
+        # giving mypy a typed handle without weakening the duck-typed signature.
+        connected_phones = cast(Any, client_pool).connected_phones
 
     if callable(connected_phones):
         try:
@@ -279,7 +281,7 @@ def connected_phones_from_pool(client_pool: object | None) -> set[str]:
             return {str(phone) for phone in phones}
 
     try:
-        clients = client_pool.clients
+        clients = getattr(client_pool, "clients", {})
     except Exception:
         clients = {}
     if isinstance(clients, dict):
@@ -330,7 +332,7 @@ def _pool_reports_connections(client_pool: object | None) -> bool:
     if callable(getattr(type(client_pool), "connected_phones", None)):
         return True
     try:
-        return isinstance(client_pool.clients, dict)
+        return isinstance(getattr(client_pool, "clients", None), dict)
     except Exception:
         return False
 
--- a/src/config.py
+++ b/src/config.py
@@ -191,7 +191,7 @@ class AppConfig(BaseModel):
     web: WebConfig = WebConfig()
     scheduler: SchedulerConfig = SchedulerConfig()
     notifications: NotificationsConfig = NotificationsConfig()
-    database: DatabaseConfig = DatabaseConfig()
+    database: DatabaseConfig = DatabaseConfig()  # type: ignore[call-arg]  # Field(gt=0) defaults are mis-seen as required
     llm: LLMConfig = LLMConfig()
     agent: AgentConfig = AgentConfig()
     security: SecurityConfig = SecurityConfig()
--- a/src/models.py
+++ b/src/models.py
@@ -667,12 +667,14 @@ class PipelineGraph(BaseModel):
         )
 
     @classmethod
-    def from_json(cls, data: str | dict) -> "PipelineGraph":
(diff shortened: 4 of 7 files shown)


axisrow:
## 🔍 Local review (cycle 1) — PR #1280

Reviewed locally (`/review` + Codex companion `sol`/`medium`), no bots pinged.
PR head: `c3884974` (rebased onto `main` `53001148`).

**Verdicts: 0 FIX, 6 SKIP (all non-blocking), 0 UNVERIFIED. Codex: APPROVE (0 findings).**

|---|---|---|---|
| SKIP | claude | `client_pool.clients` → `getattr(client_pool, "clients", {})` (and `isinstance(.clients, dict)` → `isinstance(getattr(...,None), dict)`) — identical result for all 3 cases (present+dict, present+non-dict, absent); outer try/except now dead code but behavior unchanged | _registry.py:282,333 |
| SKIP | claude | `cast(Any, client_pool).connected_phones` — cast is no-op; elif gated by `callable(type(...).connected_phones)` guarantees attr exists; identical to old getattr | _registry.py:271 |
| SKIP | claude | `from_json` `assert isinstance(data, dict)` changes exception type (AttributeError→AssertionError) for malformed non-object JSON — all realistic callers (`pipeline_service.py:590/753`, `content_pipelines.py:54`) catch `except Exception`; the only unwrapped caller (`pipeline_templates.py:41`) reads operator-seeded `template_json` that `to_json()` always emits as a dict; `-O`/`PYTHONOPTIMIZE` unused in the project so the assert is never stripped | models.py:676 |
| SKIP | claude | `# type: ignore[call-arg]` on `DatabaseConfig()` — pydantic mypy plugin mis-sees `Field(gt=0)`/`Field(ge=0)` defaults as required; real defaults (`64000`/`256`) present, `DatabaseConfig()`/`AppConfig()` construct fine at runtime | config.py:194 |
| SKIP | claude | `all_active` → `active_pipelines` rename — two sequential try-blocks reused the name with incompatible types (`list[SearchQuery]` then `list[ContentPipeline]`); pure local rename, first block's var unread after its loop | scheduler/service.py:552 |
| SKIP | claude | TYPE_CHECKING import of `numpy`/`sklearn` + field annotations (`_matrix: np.ndarray \| None`, `_index: NearestNeighbors \| None`) — runtime imports stay lazy inside `load()`/`search()` try blocks; `from __future__ import annotations` makes field annotations string-only | numpy_semantic.py |
| APPROVE | codex | "No critical ship blocker. The only runtime-observable change is that malformed non-object PipelineGraph JSON now raises AssertionError instead of the previous AttributeError; affected repository paths already catch Exception or previously failed. All other edits are runtime-neutral." 0 findings. | — |

**Triage verification (per-finding, against the code):** all 6 confirmed behavior-preserving. PR is typing-only (pyproject override, cast/getattr for duck-typed objects, type:ignore for pydantic-plugin artifact, local rename, TYPE_CHECKING lazy-import annotations).

**Local verification:** `ruff` clean; `mypy src/` 42→27 (−15); targeted `pytest` 189 passed. This is the final cycle (no FIX). Local mode does not merge — merge is yours to trigger.

PR #1371, opened 2026-08-26
chore(mypy): clear session materializer errors

axisrow:
Part of #1133.

## Summary
- clear the four platform-specific mypy errors in `src/telegram/session_materializer.py`
- preserve the Windows locking behavior; the stdlib module is missing these attributes from the active platform stubs

## Verification
- baseline on current `main`: 5 errors across 4 files
- after: 3 pre-existing errors across 3 untouched files; 0 new errors in untouched files
- `python -m mypy src --show-error-codes --no-error-summary`
- `ruff check src/telegram/session_materializer.py tests/test_session_materializer.py`
- `pytest -q tests/test_session_materializer.py` (7 passed)

Full parallel suite was started but interrupted because this checkout ran only one xdist worker and immediately produced unrelated web fixture errors.


axisrow:
## Code review — `high` effort

**Verdict: approve.** The change does what it claims, the verification in the PR body reproduces exactly, and no runtime behavior is affected. One non-blocking design note below.

### What I verified

- **Baseline reproduces exactly.** Checked the pre-PR file from `main`: 4 `attr-defined` errors, on lines 48 and 56 — `Module has no attribute "locking"`, `"LK_LOCK"`, `"LK_UNLCK"`. The two `# type: ignore[attr-defined]` comments land on precisely those two lines, and the error code matches. No over-broad bare `# type: ignore`.
- **Claim in the PR body is accurate.** `python -m mypy src` on this branch → 3 errors across 3 untouched files (`dialog_cache.py:142` `[index]`, `facade.py:905` `[union-attr]`, `collector_mixins/collection.py:1032` `[arg-type]`), 0 in `session_materializer.py`. Matches "3 pre-existing errors across 3 untouched files; 0 new errors" exactly.
- **No runtime change.** Comment-only diff. `pytest tests/test_session_materializer.py` → 7 passed; `ruff check` clean.
- **Suppression is correctly scoped.** `[attr-defined]` is the established idiom in this repo for exactly this class of issue (`mtproto_watchdog.py`, `mcp_server.py`, `deepagents_sync.py`, `pipeline_executor.py`). Nothing unusual here.

### Root cause (worth recording)

The errors are not a stub gap — typeshed's `msvcrt.pyi` gates *every* attribute behind `if sys.platform == "win32"`. The code branches on `os.name == "nt"`, which **mypy does not narrow**. `sys.platform` is mypy's only platform-narrowing predicate. So mypy analyzes the Windows branch as unreachable-but-checked on a POSIX platform and finds nothing in the module.

### Non-blocking note: a suppression-free fix exists

Because the cause is the narrowing predicate rather than a missing stub, swapping `os.name == "nt"` → `sys.platform == "win32"` (all three sites: the import guard on line 12, plus lines 41 and 54) clears all four errors with **zero** suppressions. I confirmed this — the variant type-checks clean.

Two things make that arguably the better long-term shape:

1. `sys.platform` is already the dominant idiom in `src/` (`process_control.py`, `agent_tui.py`); `os.name` appears *only* in this file.
2. It fixes the asymmetry rather than half of it. The current suppression is one-directional: CI is Linux-only, so the `fcntl` branch is the one that type-checks and `msvcrt` is suppressed. Checking the same file under `platform = win32` flips it — the `msvcrt` ignores become `unused-ignore` and four *new* `fcntl` `attr-defined` errors appear (`flock`, `LOCK_EX`, `LOCK_UN`). Real `sys.platform` narrowing makes both branches check cleanly on either platform.

**Why this is a note and not a blocker:** the repo has `warn_unused_ignores` off (`[tool.mypy]` is empty) and no Windows CI job, so the stale-ignore scenario cannot actually fire today. The ignores are honest, minimal, and correctly coded. For a scoped `#1133` cleanup PR, taking the two-line suppression over a three-site refactor of live locking code is a reasonable trade — the refactor touches the cross-process lock path and deserves its own PR with its own justification, not a drive-by in a mypy sweep.

Suggest filing the `os.name` → `sys.platform` swap as a follow-up under #1133 rather than expanding scope here.

### Note on the PR description

"the stdlib module is missing these attributes from the active platform stubs" is slightly off — the attributes *are* in the stubs, just behind a `sys.platform` guard the code doesn't trip. Worth correcting so the next reader doesn't go hunting for a typeshed bug. Cosmetic only.

PR #1375, opened 2026-08-26
chore(mypy): use platform-aware session lock guards

axisrow:
Part of #1133.

## Summary
- use `sys.platform == "win32"` for all three platform guards in `SessionMaterializer`
- remove the two targeted mypy ignores; both Windows and POSIX branches now use the platform predicate understood by mypy
- preserve the existing lock implementation and runtime behavior

## Verification
- baseline on current `main` after #1371: 3 mypy errors in 3 files
- after: same 3 errors in the 3 untouched files; `session_materializer.py` is clean
- `ruff check src/telegram/session_materializer.py tests/test_session_materializer.py`
- `pytest -q tests/test_session_materializer.py` (7 passed)
- `git diff --check`


axisrow:
## Code review — `/code-review` (high effort)

**No findings.** The diff is clean; nothing to report.

Scope reviewed: `gh pr diff 1375` — `src/telegram/session_materializer.py` only. Three `os.name == "nt"` → `sys.platform == "win32"` guard swaps, removal of the two `# type: ignore[attr-defined]` comments added by #1371, and dropping the now-unused `os` import.

What I checked and why it holds:

- **All three guards swapped consistently** (module import guard, `acquire`, `release`). No mixed state where the module imports `msvcrt` but takes the `fcntl` branch, or where acquire/release disagree — that asymmetry was the real bug risk in this change.
- **Behaviorally identical on every platform CPython supports.** Windows: both predicates true. Cygwin/MSYS: both false (`os.name == "posix"`, `sys.platform == "cygwin"`), so both route to `fcntl`.
- **The removed ignores were redundant, not masking errors.** `sys.platform` is the predicate mypy narrows on, so the Windows branch now type-checks against the real `msvcrt` stubs. `mypy src/telegram/session_materializer.py` → clean. Ignores were removed rather than added, so `warn_unused_ignores` cannot regress.
- **`os` import removal is safe** — no remaining `os` references in the file; it was used only by the deleted guards.
- **No test depended on the old predicate** — zero `os.name` hits across `src/` and `tests/`, so nothing was monkeypatching `src.telegram.session_materializer.os.name`. That silent loss of a patch target was the main latent hazard for this class of change, and it doesn't apply here.
- **Runtime verification**: `tests/test_session_materializer.py` 7 passed; `test_regression_infra_coverage.py` + `test_main_client_pool_collector_paths.py` 174 passed; `ruff check` clean.

Scope note (not a defect): a local `git diff main...HEAD` is much larger (CLAUDE.md, docs) because the branch carries already-merged upstream commits; those are not part of this PR.

PR #1377, opened 2026-08-26
chore(mypy): narrow dialog cache aggregate row

axisrow:
Part of #1133.

## Summary
- make the aggregate-row invariant explicit in `DialogCacheRepository.upsert_dialogs`
- document that `MAX()` without `GROUP BY` returns one row even for an empty snapshot
- preserve runtime behavior while clearing the final `index` error in this module

## Verification
- baseline on current `main` after #1376: 2 errors in 2 files
- after: 1 error in untouched `collector_mixins/collection.py`; `dialog_cache.py` is clean
- `ruff check src/database/repositories/dialog_cache.py tests/repositories/test_dialog_cache_repository.py`
- `pytest -q tests/repositories/test_dialog_cache_repository.py` (5 passed)
- `git diff --check`


axisrow:
@claude review

PR #1378, opened 2026-08-26
chore(mypy): type collector resolve protocol

axisrow:
Part of #1133.

## Summary
- type the duck-typed collector dependency in `collector_resolve.py` with the same structural protocol used by the collector mixins
- clear the collector mixin delegation `arg-type` error without changing runtime behavior

## Verification
- baseline on current `main` after #1377: 2 errors in 2 files
- after: 1 error in untouched `database/facade.py`; collector resolve and mixin are clean
- `ruff check src/telegram/collector_resolve.py src/telegram/collector_mixins/collection.py`
- `python -m mypy src --show-error-codes --no-error-summary`
- focused collector resolve tests passed
- `git diff --check`


axisrow:
@claude review

